PR #98, opened 2025-05-14
feat: add support for managing projects

scotwells:
Introduces support for managing projects within the resource manager service. Projects will be used by organizations to organize resources they create within the system.


internal/grpc/server/projects_service.go
```diff
@@ -0,0 +1,168 @@
...
+		projectToUpdate := proto.Clone(existingProject).(*resourcemanagerpb.Project)
+		fmutils.Overwrite(projectFromRequest, projectToUpdate, req.GetUpdateMask().GetPaths())
+
+		if errs := validation.ValidateProject(projectToUpdate); len(errs) > 0 {
+			return nil, errs.GRPCStatus().Err()
+		}
+
+		projectToUpdate.UpdateTime = timestamppb.Now()
+		return projectToUpdate, nil
+	}
```
JoseSzycho, line 103: On the resource definition, it mentions that the annotations object would be preserved when modifying the resource.  I think that there the annotations would be overwritten

https://github.com/datum-cloud/milo/blob/16bdb58b71ed2f3faf6b8d45cf2411214551bcca/api/datum/resourcemanager/v1alpha/projects_resources.proto#L65

scotwells: @JoseSzycho the annotations would be altered based on the values in the update mask. This way the user can remove an annotation if they'd like to. I can adjust the documentation on the annotations field so that's more clear.

JoseSzycho: It's clear to me now! I just wanted to be sure

JoseSzycho (approved):

internal/grpc/server/projects_service.go
scotwells: Good call out!

config/services/resourcemanager.datumapis.com.yaml
```diff
@@ -15,3 +15,17 @@ spec:
...
+    - list
+    - get
+    - create
+    - update
+    - delete
+    - move
+    resourceNamePatterns:
+    - "projects/{project}"
+    - "organizations/{organization}/projects/{project}"
+    - "folders/{folder}/projects/{project}"
```
mustela, line 31: What are folders?

internal/grpc/server/projects_service.go
```diff
@@ -0,0 +1,168 @@
...
+		return nil, errs.GRPCStatus().Err()
+	}
+
+	if req.GetValidateOnly() {
+		metadata := &resourcemanagerpb.CreateProjectMetadata{}
+		return longrunning.ResponseOperation(metadata, req.GetProject(), true)
+	}
+
+	project.Name = fmt.Sprintf("projects/%s", project.ProjectId)
+	project.Parent = req.GetParent()
```
mustela, line 40: I guess verifying that the user has access to the parent is happening outside this module, right?

scotwells: That's correct. We use a gRPC interceptor (basically a middleware) to do [request authorization based on the authenticated subject](https://github.com/datum-cloud/milo/blob/main/internal/grpc/auth/interceptor.go#L32-L59), the resource they're trying to access, and the method they're accessing. 

This interceptor is leveraging the `datum.api.required_permissions` protobuf annotation to determine what permission is required to access the endpoint. The interceptor will automatically identify the resource based on other metadata included in the request or annotations on the protobuf definition. 

```
option (datum.api.required_permissions) = "iam.datumapis.com/serviceAccounts.update";
```

The [ResourceNameResolver](https://github.com/datum-cloud/milo/blob/main/internal/grpc/auth/interceptor.go#L208) function goes through how the interceptor identifies the resource name for a request.

PR #99, opened 2025-05-14
Automatically resolve resource parents

scotwells:
This change modifies the subject extractor to resolve the subject name in the format`user:<email>` so it follows the correct subject name format that can be used with the `subject.Parse` function. The `SubjectExtractor` interface was moved to the `subject` package so it's coupled with the other subject related code. A new `SubjectReference` type was returned from the `SubjectResolver` interface so we can include additional metadata around the subject that was resolved.

The AccessCheck endpoint now also resolves the parent resources for the access check request so the context doesn't need to be provided by the client. The AccessCheck endpoint will also now do subject resolution so clients calling AccessCheck don't need to understand how to resolve the subject name.

### DIFF
--- a/cmd/apiserver/app/add-resources.go
+++ b/cmd/apiserver/app/add-resources.go
@@ -94,13 +94,18 @@ to the system.
 				RoleStorage: roleStorage,
 			}
 
+			subjectResolver, err := subject.DatabaseResolver(db)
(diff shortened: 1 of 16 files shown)


JoseSzycho (approved):
LGTM

PR #128, opened 2025-05-19
Add functionality to load users IAM service

JoseSzycho:
This PR introduces functionality to add users into the IAM service, enabling us to bootstrap the system with initial user accounts. It also includes a validation step to ensure email address ownership consistency.

When adding a new user, the system checks if an existing user with the same email address already exists. If the new user’s account resource name does not match the existing user’s resource name, an error is thrown. This prevents accidental reassignment of email addresses to different accounts.


scotwells:
This command is no longer needed now that we're using the k8s apiserver library. We can use kubectl or other Kubernetes API compatible tooling (e.g. FluxCD) to load resources at deployment time.

PR #132, opened 2025-05-20
Fix JWT Authentication Failure Due to Improper Bearer Token Extraction

JoseSzycho:
This PR fixes an issue where all authentication attempts were incorrectly returning HTTP 401 Unauthorized, even when a valid token was provided or when the token had insufficient permissions.

The root cause was traced to the extractBearerToken function, which was incorrectly returning the entire Authorization header value (i.e., Bearer <token>) instead of extracting just the token part. This led to a failure during JWT verification.

The following error was consistently logged by the server:

`{"time":"2025-05-20T14:34:08.394476333Z","level":"ERROR","msg":"failed to verify JWT bearer token against authentication provider","error":"failed to parse jws: failed to decode protected headers: failed to decode source: illegal base64 data at input byte 6","authentication_provider":"http://localhost:8082/oauth/v2/keys"}`

The fix ensures that the Bearer prefix is properly removed before passing the token to the JWT verification logic.

Reference: [extractBearerToken function](https://github.com/datum-cloud/milo/blob/872fa829b408d5ed11ad1964ed572156d92da625/internal/grpc/auth/jwt/subject_extractor.go#L85)

### DIFF
--- a/internal/grpc/auth/jwt/subject_extractor.go
+++ b/internal/grpc/auth/jwt/subject_extractor.go
@@ -102,5 +102,6 @@ func extractBearerToken(ctx context.Context) (string, error) {
 		return "", fmt.Errorf("invalid authorization header format")
 	}
 
-	return parts[1], nil // Return the token part
+	tokenPart := strings.Split(parts[1], "Bearer")[1]
+	return tokenPart, nil // Return the token part
 }


internal/grpc/auth/jwt/subject_extractor.go
```diff
@@ -102,5 +102,6 @@ func extractBearerToken(ctx context.Context) (string, error) {
 		return "", fmt.Errorf("invalid authorization header format")
 	}
 
-	return parts[1], nil // Return the token part
+	tokenPart := strings.Split(parts[1], "Bearer")[1]
+	return tokenPart, nil // Return the token part
```
scotwells, line 106: Wouldn't `parts[1]` here be the token? I would expect `parts[0]` to be `Bearer` since it's splitting the auth header value by an empty space. 

Can you provide an example curl request that replicates the issue you were running into?

JoseSzycho: @scotwells Thank you for your reply, I realized that I got a bad update from Postman. Postman is adding a double Bearer prefix on all of the requests, and that was the cause of the error

PR #182, opened 2025-07-08
Adding User Preferences CRD

mustela:
This PR introduces a new UserPreferences Custom Resource Definition (CRD) to the service. It will be used to store per-user settings such as timezone and theme preference. This lays the groundwork for supporting personalized user experiences in future features.


pkg/apis/iam/v1alpha1/userpreference_types.go
```diff
@@ -0,0 +1,57 @@
...
+// +kubebuilder:printcolumn:name="Theme",type="string",JSONPath=".spec.theme"
+// +kubebuilder:printcolumn:name="Ready",type="string",JSONPath=".status.conditions[?(@.type=='Ready')].status"
+// +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
+// +kubebuilder:resource:path=userpreferences,scope=Cluster
+type UserPreference struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   UserPreferenceSpec   `json:"spec,omitempty"`
+	Status UserPreferenceStatus `json:"status,omitempty"`
```
joshlreese, line 22: @mustela / @scotwells - Do you think this is a situation where it'd make sense to omit the `spec` and `status` fields, or is there a possibility that these settings may not apply until a reconciler has processed them?

From https://github.com/robscott/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status:

> All objects that represent a physical resource whose state may vary from the user's desired intent SHOULD have a `spec` and a `status`. Objects whose state cannot vary from the user's desired intent MAY have only `spec`, and MAY rename `spec` to a more appropriate name.

This would be similar to how a `ConfigMap` or `Secret` have `data` fields, and no `spec` or `status`.

pkg/apis/iam/v1alpha1/userpreference_types.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+// UserPreferenceSpec defines the desired state of UserPreference
+type UserPreferenceSpec struct {
+	// Reference to the user these preferences belong to.
+	// +kubebuilder:validation:Required
+	UserRef ScopedRoleReference `json:"userRef"`
```
joshlreese, line 29: Was it intentional to use `ScopedRoleReference` when referencing a user here?

pkg/apis/iam/v1alpha1/userpreference_types.go
```diff
@@ -0,0 +1,57 @@
...
+
+	// The user's timezone preference (e.g., "America/New_York", "Europe/London").
+	// +kubebuilder:validation:Optional
+	Timezone string `json:"timezone,omitempty"`
```
joshlreese, line 33: Is there a default for this? Also, should we note that this will only be used by the portal, and not API responses?

pkg/apis/iam/v1alpha1/userpreference_types.go
scotwells: @joshlreese I think a status could still make sense here since some of these settings might need to be reconciled against other systems like Zitadel.

scotwells:
Are we thinking that a UserPreference resource would automatically be created for the user when the user is created in the system?

pkg/apis/iam/v1alpha1/userpreference_types.go
mustela: Ill add defaults, good call!

What do you mean with "not API responses"?

joshlreese: > What do you mean with "not API responses"?

Timestamps returned by the API will not be influenced by this setting, and will continue to be in UTC. Maybe renaming this to `DashboardTimezone` might make sense?

As for the default, I suspect it would be safe to say that it defaults to the timezone of the user's browser, which seems to be typical.

mustela: Oh, I see now. But I wouldn’t use DashboardTimezone, because this attribute could be used for other things as well—like notifications, for instance.

mustela:
@scotwells @scotwells would you mind to take another look?

pkg/apis/iam/v1alpha1/userpreference_types.go
scotwells: @mustela how would this option be used for something like notifications? Do you mean when viewing notifications in the portal?

mustela:
@scotwells are there any other resources I can check to understand how to do that?

scotwells:
@mustela the [user webhook](https://github.com/datum-cloud/milo/blob/main/internal/webhooks/iam/v1alpha1/user_webhook.go#L44) is an example of automatically creating a resource when another resource is created. We're using [validating webhooks](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/) right now, but hope to use [Kyverno](https://kyverno.io) in the future. This webhook is creating a policy binding that binds a role to the user that's being created so they have access to manage themselves.

We'd use a validating webhook to add any complex validation that needs to be performed (e.g. the validating webhook needs to query the API).

internal/webhooks/iam/v1alpha1/user_webhook.go
```diff
@@ -109,3 +114,28 @@ func (v *UserValidator) createSelfManagePolicyBinding(ctx context.Context, user
 
 	return nil
 }
+
+// createUserPreference creates a UserPreference for the new user
+func (v *UserValidator) createUserPreference(ctx context.Context, user *iamv1alpha1.User) error {
```
scotwells, line 119: You'll want to create a new PolicyBinding to grant the user access to get / update this UserPreference resource.

mustela: Updated @scotwells

scotwells: @mustela you're still missing the creation of the Policy Binding resource after the UserPreference resource is created by the webhook. The policy binding is what would grant the user access to retrieve / update the UserPreference object.

Also just realized you're missing a ProtectedResource definition for the UserPreference CRD so the IAM system will be programmed to support the permissions available. You can see an [example here](https://github.com/datum-cloud/milo/blob/main/config/protected-resources/iam/role.yaml).

mustela: @scotwells updated!

pkg/apis/iam/v1alpha1/userpreference_types.go
scotwells: I agree it shouldn't be tool dependent, I just don't want to potentially cause confusion that this timezone setting would be respected or used for backend API responses.

config/samples/iam/v1alpha1/userpreference-policybinding-example.yaml
```diff
@@ -0,0 +1,30 @@
...
+---
+apiVersion: iam.miloapis.com/v1alpha1
+kind: UserPreference
+metadata:
+  name: userpreference-john-doe
+  namespace: milo-system
+spec:
+  userRef:
+    name: "john-doe"
+  timezone: "America/New_York"
```
scotwells, line 29: Should update the sample to remove the timezone since that's not on the spec anymore.

scotwells (dismissed):

config/samples/iam/v1alpha1/userpreference-policybinding-example.yaml
mustela: oh missed this one. Fixed 🙇🏼

scotwells (approved):

PR #184, opened 2025-07-10
Create control plane overlays

scotwells:
The controller manager deployed to the core control plane needs RBAC resources that are specific to the core control plane. This is because the controller when running in the core control plane needs to be able to access the ProjectControlPlane resource in the infra cluster.

This PR creates overlays for the controller manager that supports deploying it for the core control plane and project control planes.

This also includes a minor bug fix that ensures the `spec` is required when creating an Organization. This addresses an issue where the organization type could be empty.


joshlreese (approved):
Seems like there's a few unrelated changes - assuming those are intentional, this looks good.

PR #185, opened 2025-07-10
Fix organization membership sync

scotwells:
This resolves an issue where the OrganizationMembership resource wasn't being updated when an Organization or User was updated. This was caused by the field indexers not being registered for the field selectors used in the watch handlers for the Organization and User resources. 

### Other Changes

- Fixed issue with RBAC configuration being created in the wrong spot after it was moved in #184
- Adds more printer columns to the Organization type


brandon-burciaga (approved):
tests are failing, but the code looks good to me ✅

PR #187, opened 2025-07-15
feat(quota): adds go type files and crds

brandon-burciaga:
- Implementation of the phase one MVP quota management system
- _**Draft while controllers are tested**_ - will convert and add reviewers when ready for full review


Taskfile.yaml
```diff
@@ -45,6 +45,153 @@ tasks:
...
+      # Create the resource claim for 1 project for the organization
+      - kubectl apply -f config/samples/quota/v1alpha1/resourceclaim_project_for_org.yaml
+      # Create the resource claim for 2 additional projects for the organization
+      - kubectl apply -f config/samples/quota/v1alpha1/resourceclaim_projects_for_org_additional.yaml
+      # Create the resource claim for 1 additional project for the organization
+      - kubectl apply -f config/samples/quota/v1alpha1/resourceclaim_projects_for_org_additional_2.yaml
+      # Create the resource claim for 2 additional HTTPProxy instances for the project
+      - kubectl apply -f config/samples/quota/v1alpha1/resourceclaim_httpproxies_per_project.yaml
+
+  kind-create-all-quota-resources:
```
brandon-burciaga, line 156: This should be run after `task kind-setup-and-run`. To simplify for debugging local webhook triggering, quota resource creation is commented out (other than registrations), and organization creation is the last command run since this is what should be triggering the organization webhook.

config/dev/webhook_patch.yaml
```diff
@@ -26,6 +26,11 @@ patch: |-
     from: /webhooks/1/clientConfig/service/path
   - op: remove
     path: /webhooks/1/clientConfig/service
+  - op: move
```
brandon-burciaga, line 29: This was added when trying to get the webhooks to be triggered locally

config/samples/resourcemanager/v1alpha1/organization.yaml
```diff
@@ -3,4 +3,4 @@ kind: Organization
 metadata:
   name: datum
 spec:
-  type: Business
+  type: Personal
```
brandon-burciaga, line 6: Updated since `Business` is no longer supported.

config/webhook/manifests.yaml
```diff
@@ -4,6 +4,48 @@ kind: MutatingWebhookConfiguration
 metadata:
   name: resourcemanager.miloapis.com
 webhooks:
+- admissionReviewVersions:
```
brandon-burciaga, line 7: Configuration for the mutating resourceclaim creation webhook

Taskfile.yaml
```diff
@@ -45,6 +45,153 @@ tasks:
...
+
+      # 8. Apply the development webhook manifests - necessary??
+      # - kubectl apply -k config/dev/
+
+      # 9. Run the controller manager locally
+      - |
+        ./bin/milo controller-manager \
+          --leader-elect=false \
+          --authentication-skip-lookup \
+          --secure-port=43271 \
```
brandon-burciaga, line 94: Perhaps this port is wrong? Also, should I be using `--controller-runtime-webhook-port=9443` ?

Taskfile.yaml
```diff
@@ -45,6 +45,153 @@ tasks:
...
+          echo "Port 43271 in use – killing existing process..."
+          kill -9 $(lsof -i tcp:43271 -t)
+        fi
+        if lsof -i tcp:9443 -t >/dev/null 2>&1; then
+          echo "Port 9443 in use – killing existing process..."
+          kill -9 $(lsof -i tcp:9443 -t)
+        fi
+
+      # 8. Apply the development webhook manifests - necessary??
+      # - kubectl apply -k config/dev/
```
brandon-burciaga, line 87: Not sure if this needs to be applied, but while debugging in cursor it was suggested that it should be.

brandon-burciaga (commented):
Left a few comments regarding getting the webhooks to trigger and log locally.

The `dev/webhook_patch.yaml` file uses `prefix: "https://host.docker.internal:9443"`, but since I am not running this locally in docker, it may be incorrect?

https://github.com/datum-cloud/milo/blob/cffdb83aacddf107b2b861ff42ad0e9cad98f1f4/config/dev/webhook_patch.yaml#L41

The ` milo-system` namespace is also not being created - so quota resources other than registrations are failing to be applied. I am assuming this is because I am running the controller manager in the core control plane and not project control plane...?

config/webhook/manifests.yaml
```diff
@@ -27,6 +27,40 @@ webhooks:
   sideEffects: NoneOnDryRun
 ---
 apiVersion: admissionregistration.k8s.io/v1
+kind: MutatingWebhookConfiguration
```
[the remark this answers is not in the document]
scotwells: Since you're registering a webhook directly with the system, you'll want to remove the kubebuilder annotations. The webhook configuration will need to be managed based on the resources that are protected by the quota system.

PR #197, opened 2025-07-18
Add UserDeactivation CRD and user account activation state management

JoseSzycho:
## Summary

This PR introduces the ability to activate and deactivate user accounts in Milo by adding a new `UserDeactivation` Custom Resource Definition (CRD) and extending the `User` resource to support account state management.

## Key Changes

- **Add UserDeactivation CRD**
  - Introduces a new cluster-scoped `UserDeactivation` resource for managing user account deactivation.
  - Includes fields for referencing the user, specifying the reason, description, and who initiated the deactivation.
  - Adds status conditions for tracking reconciliation state.

- **Extend User CRD for Activation State**
  - Adds a `state` field to `UserStatus` to represent whether a user is `Active` or `Inactive`.
  - The `state` field is managed exclusively by the `UserDeactivation` resource and cannot be changed directly by users.
  - Updates CRD OpenAPI schema and kubebuilder print columns to reflect the new state field.

- **Register New Types**
  - Registers `UserDeactivation` and `UserDeactivationList` types in the API scheme.

- **Autogenerate CRDs and Deepcopy Methods**
  - Updates generated CRDs and deepcopy methods to support the new resource and fields.

## Motivation

These changes enable administrators to programmatically deactivate and reactivate user accounts, improving security and lifecycle management. Deactivation revokes all existing sessions and prevents authentication until reactivation.

## Implementation Details

- Creating a `UserDeactivation` resource for a user sets their state to `Inactive` and disables authentication.
- Deleting the `UserDeactivation` resource reactivates the user.
- The new CRD and fields are fully integrated into the API and code generation.

>[!IMPORTANT]
> The update of the User State, and the deactivation/reactivation of the user is handled on the auth-provider controllers side.


scotwells:
@JoseSzycho looks great. 🚀 

We should add a webhook to validate that the user reference actually exists when creating the User Deactivation resource. 

If you want to iterate on that in multiple PRs, let's create an integration branch (e.g. `integration/user-deactivation`) so we can integrate all changes there and test them before we merge into the main branch.

scotwells:
@JoseSzycho the webhook validation makes sense in this case because we will only want a UserDeactivation resource to exist if the user it’s referencing exists.

scotwells:
@JoseSzycho good call, the webhook is the perfect place for that. You can use the admission package to retrieve the request info from the context which will include the authenticated user info. There’s some examples in the existing webhooks

https://github.com/datum-cloud/milo/blob/main/internal/webhooks/resourcemanager/v1alpha1/organization_webhook.go#L51C24-L51C42

JoseSzycho:
@scotwells Thanks for the example! 

Starting to work based on this deactivation flow: enhancement PR https://github.com/datum-cloud/enhancements/pull/218

PR #201, opened 2025-07-21
bugfix: use correct subject name in project webhook

scotwells:
The subject name was being set to the username of the user which is not the correct name. The correct name to use is the metadata name of the user resource.


brandon-burciaga (approved):
✅

PR #205, opened 2025-07-22
Defining Milo SLOs w/ Alerting

OscarLlamas6:


config/telemetry/milo-slo/prometheus-rules.yaml
```diff
@@ -0,0 +1,30 @@
...
+          labels:
+            severity: critical
+            slo_violation: "true"
+          annotations:
+            summary: "Project {{ $labels.name }} is stuck creating for over 60 seconds"
+            description: "Project {{ $labels.name }} has been in creation state for {{ $value }} seconds without reaching Ready status, which exceeds the 60-second SLO threshold."
+
+        - alert: ProjectReadyStatusSLOViolation
+          expr: |
+            (sum by (name, namespace) (milo_projects_status_condition_last_transition_time{type="Ready"}) - sum by (name, namespace) (milo_projects_created_timestamp)) > 60
```
scotwells, line 23: @OscarLlamas6 what's the goal of this alert? It seems like this one would only start firing after the project has reached a ready state unless I'm misinterpreting this query.

OscarLlamas6: @scotwells You're not wrong, it's exactly that, initially I only had this one and then I thought "we want to know when the project takes more than 60 seconds to create regardless of whether it's in progress or finished" and just as you say this alert would only be activated when the project reaches Ready, so if a user is experiencing degradation in project creation, our alert would notify us only if their project after X time was successful. It's part of what I wanted to talk about in today's session but I'm glad you clarified it here.

PR #208, opened 2025-07-28
Implement and Test UserDeactivation Validation Webhook with Defaulting and Enhanced Logic

JoseSzycho:
This PR introduces the following changes related to the `UserDeactivation` resource:

- **Webhook Implementation**
  - Added the UserDeactivation validation webhook and integrated it into the controller manager.
  - Implemented the `UserDeactivationMutator` to default the `deactivatedBy` field and enhance validation logic.

- **Validation Logic**
  - During validation, the webhook checks that the referenced user exists; otherwise, the `UserDeactivation` resource cannot be created.
  - Ensures that each user can only have one `UserDeactivation` resource. If a user already has a deactivation, additional ones are rejected.

- **Defaulting**
  - Added a Defaulter to automatically populate the `deactivatedBy` field if it is not set.

- **Testing**
  - Added comprehensive unit tests for the UserDeactivation validation logic.
  - Enhanced validation tests with additional scenarios and defaulting behavior to ensure robustness.

- **Manifests**
  - Updated relevant manifests to support the new webhook functionality.

### DIFF
--- a/internal/webhooks/iam/v1alpha1/userdeactivation_webhook.go
+++ b/internal/webhooks/iam/v1alpha1/userdeactivation_webhook.go
@@ -17,8 +17,6 @@ import (
 
 var userdeactivationlog = logf.Log.WithName("userdeactivation-resource")
 
-// +kubebuilder:webhook:path=/validate-iam-miloapis-com-v1alpha1-userdeactivation,mutating=false,failurePolicy=fail,sideEffects=NoneOnDryRun,groups=iam.miloapis.com,resources=userdeactivations,verbs=create,versions=v1alpha1,name=vuserdeactivation.iam.miloapis.com,admissionReviewVersions={v1,v1beta1},serviceName=milo-controller-manager,servicePort=9443,serviceNamespace=milo-system
-
 func SetupUserDeactivationWebhooksWithManager(mgr ctrl.Manager, systemNamespace string) error {
 	userdeactivationlog.Info("Setting up iam.miloapis.com userdeactivation webhooks")
 
@@ -31,6 +29,8 @@ func SetupUserDeactivationWebhooksWithManager(mgr ctrl.Manager, systemNamespace 
 		}).
 		Complete()
 }
+
+// +kubebuilder:webhook:path=/mutate-iam-miloapis-com-v1alpha1-userdeactivation,mutating=true,failurePolicy=fail,sideEffects=None,groups=iam.miloapis.com,resources=userdeactivations,verbs=create,versions=v1alpha1,name=muserdeactivation.iam.miloapis.com,admissionReviewVersions={v1,v1beta1},serviceName=milo-controller-manager,servicePort=9443,serviceNamespace=milo-system
 
 // UserDeactivationMutator sets default values on UserDeactivation resources.
 type UserDeactivationMutator struct{}
@@ -56,6 +56,8 @@ func (m *UserDeactivationMutator) Default(ctx context.Context, obj runtime.Objec
 
 	return nil
 }
+
+// +kubebuilder:webhook:path=/validate-iam-miloapis-com-v1alpha1-userdeactivation,mutating=false,failurePolicy=fail,sideEffects=None,groups=iam.miloapis.com,resources=userdeactivations,verbs=create,versions=v1alpha1,name=vuserdeactivation.iam.miloapis.com,admissionReviewVersions={v1,v1beta1},serviceName=milo-controller-manager,servicePort=9443,serviceNamespace=milo-system
 
 type UserDeactivationValidator struct {
 	client          client.Client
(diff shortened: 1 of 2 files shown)


internal/webhooks/iam/v1alpha1/userdeactivation_webhook.go
```diff
@@ -0,0 +1,126 @@
...
+	if !ok {
+		return fmt.Errorf("failed to cast object to UserDeactivation")
+	}
+	userdeactivationlog.Info("Defaulting UserDeactivation", "name", ud.GetName())
+
+	// If deactivatedBy is already set, respect the provided value.
+	if ud.Spec.DeactivatedBy != "" {
+		userdeactivationlog.Info("DeactivatedBy already set, skipping defaulting", "name", ud.GetName())
+		return nil
+	}
```
scotwells, line 49: I'd recommend we prevent this field from being user-specified. It should always be resolved based on the authenticated user.

internal/webhooks/iam/v1alpha1/userdeactivation_webhook.go
```diff
@@ -0,0 +1,126 @@
...
+
+	return nil
+}
+
+type UserDeactivationValidator struct {
+	client          client.Client
+	systemNamespace string
+}
+
+func (v *UserDeactivationValidator) ValidateCreate(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
```
scotwells, line 70: I'd recommend building up a list of validation errors and using [the `errors.NewInvalid()` function](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors#NewInvalid) to return a list of invalid errors back to the client. This will ensure the caller has a list of every invalid field in the request instead of having to call the API multiple times to resolve invalid errors.

internal/webhooks/iam/v1alpha1/userdeactivation_webhook.go
```diff
@@ -0,0 +1,126 @@
...
+	if err := v.client.List(ctx, &existingUDList); err != nil {
+		userdeactivationlog.Error(err, "failed to list existing UserDeactivations", "userName", userName)
+		return nil, fmt.Errorf("failed to list existing UserDeactivations for user '%s': %w", userName, err)
+	}
+	for _, existing := range existingUDList.Items {
+		if existing.Spec.UserRef.Name == userName && existing.DeletionTimestamp == nil {
+			userdeactivationlog.Error(fmt.Errorf("a UserDeactivation already exists for user '%s'", userName), "existing UserDeactivation", "name", existing.Name)
+			return nil, fmt.Errorf("a UserDeactivation already exists for user '%s'", userName)
+		}
+	}
```
scotwells, line 113: This should result in [an Already Exists error](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors#NewAlreadyExists).

scotwells (commented):
Great work, just a few minor comments.

internal/webhooks/iam/v1alpha1/userdeactivation_webhook.go
JoseSzycho: Thanks for the fast review. Fixed!

internal/webhooks/iam/v1alpha1/userdeactivation_webhook.go
JoseSzycho: Done, thanks!

internal/webhooks/iam/v1alpha1/userdeactivation_webhook.go
```diff
@@ -0,0 +1,116 @@
...
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	iamv1alpha1 "go.miloapis.com/milo/pkg/apis/iam/v1alpha1"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+var userdeactivationlog = logf.Log.WithName("userdeactivation-resource")
+
+// +kubebuilder:webhook:path=/validate-iam-miloapis-com-v1alpha1-userdeactivation,mutating=false,failurePolicy=fail,sideEffects=NoneOnDryRun,groups=iam.miloapis.com,resources=userdeactivations,verbs=create,versions=v1alpha1,name=vuserdeactivation.iam.miloapis.com,admissionReviewVersions={v1,v1beta1},serviceName=milo-controller-manager,servicePort=9443,serviceNamespace=milo-system
```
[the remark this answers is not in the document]
JoseSzycho: Thanks for the article. It's quite interesting. I'll take the rest of the day studying side-effects and other configurations I found on the article.

Additionally, I added the mutation scaffolding

scotwells (approved):

PR #209, opened 2025-07-29
Introduce UserDeactivation Resource, Extend User CDR with Status, and Add Local Testing Docs for Authentication Webhook

JoseSzycho:
### Summary

This PR introduces the following changes:

1.  **New UserDeactivation Resource**
    
    *   Added a new resource to handle user deactivation events.
        
    *   Supports tracking and managing the lifecycle of user deactivation.
        
2.  **Update to User CDR**
    
    *   The User CDR has been extended with a new field to expose the **current user status**.
        
    *   This enables downstream services to easily consume and act upon the user’s active/deactivated state.
        
3.  **Documentation for Local Testing**
    
    *   Added detailed documentation on how to test the **authentication webhook** locally.
        
    *   Includes prerequisites, configuration steps, and example test flows.


scotwells:
> @scotwells - I'm not really sure what you're proposing here regarding "return the user information" - what would be returned? Or are you proposing that Zitadel can determine if a token has been revoked, and be added in the critical path of every API request?

I was proposing using [Webhook Token Authentication](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#webhook-token-authentication) to validate the token against Zitadel. The TokenReview API requires the user information be returned in the status. 

```yaml
{
  "apiVersion": "authentication.k8s.io/v1",
  "kind": "TokenReview",
  "status": {
    "authenticated": true,
    "user": {
      # Required
      "username": "janedoe@example.com",
      # Optional
      "uid": "42",
      ...
  }
}
```

We'd be able to avoid lookups on every API request by configuring the `--authentication-token-webhook-cache-ttl` option with an acceptable TTL. By default it's set to 2 mins.

joshlreese:
@scotwells - We spoke on slack to clarify that you're recommending the webhook as an alternative to the filter, it wasn't clear to me if it was in addition to or a part of what's being added here.

I like this more than the filter and validating policy approach. My main concern is that this places a dependency on Zitadel. Regardless of caching, can we make sure it's deployed with redundant API servers?

JoseSzycho:
@scotwells @joshlreese Just thinking—I like the idea of using a `ValidatingAdmissionPolicy`.  

Wouldn’t it be more efficient to investigate whether we could leverage `auth-provider-openfga` with a configuration that prevents users from getting authorization once their account has been deactivated?

joshlreese:
> Just thinking—I like the idea of using a ValidatingAdmissionPolicy.

I think we should explore the approach that @scotwells mentioned before looking into ValidatingAdmissionPolicies.

> Wouldn’t it be more efficient to investigate whether we could leverage `auth-provider-openfga` with a configuration that prevents users from getting authorization once their account has been deactivated?

I assumed that preventing issuing tokens to deactivated users was already addressed, is it not?

JoseSzycho:
@scotwells @joshlreese Continuing the work on this PR https://github.com/datum-cloud/auth-provider-zitadel/pull/18

scotwells:
@JoseSzycho is this PR still needed?

JoseSzycho:
@scotwells It is, as it contains the UserDeactivation CDR, and an update to the User CDR

PR #222, opened 2025-08-08
feat: Implement UserController for managing user states and finalization

JoseSzycho:


internal/controllers/iam/user_controller.go
```diff
@@ -0,0 +1,180 @@
...
+// This is used to clean up associated UserDeactivations when a User is deleted.
+type userFinalizer struct {
+	Client client.Client
+}
+
+// Finalize implements the finalizer.Finalizer interface.
+// TODO: Implement this
+func (f *userFinalizer) Finalize(ctx context.Context, obj client.Object) (finalizer.Result, error) {
+	return finalizer.Result{}, nil
+}
```
scotwells, line 42: We can remove the finalizer stuff since it's not used by this controller.

internal/controllers/iam/user_controller.go
```diff
@@ -0,0 +1,180 @@
...
+
+	// Defining the desired user state
+	var desiredState iamv1alpha1.UserState
+	// If there are any UserDeactivations for this user, the user is inactive
+	if len(udList.Items) > 0 {
+		desiredState = iamv1alpha1.UserStateInactive
+	} else {
+		desiredState = iamv1alpha1.UserStateActive
+	}
+	user.Status.State = desiredState
```
scotwells, line 100: We should set this based on the status of the UserDeactivation. If the UserDeactivation hasn't been processed yet, then the user will still be active.

internal/webhooks/iam/v1alpha1/userdeactivation_webhook.go
```diff
@@ -54,6 +57,23 @@ func (m *UserDeactivationMutator) Default(ctx context.Context, obj runtime.Objec
...
+	}
+
+	ud.OwnerReferences = []metav1.OwnerReference{
+		{
+			APIVersion: iamv1alpha1.SchemeGroupVersion.String(),
+			Kind:       "User",
+			Name:       user.Name,
+			UID:        user.GetUID(),
+		},
+	}
```
scotwells, line 75: You can simplify this by using the `controllerutil.SetOwnerReference()` function. 

```suggestion
if err := SetOwnerReference(user, ud, m.scheme); err != nil {
     return errors.NewInternalError(fmt.Errorf("failed to set owner reference for user deactivation: %w", errr))
}
```

You'll want to add a new `scheme` field to the webhook and use `mgr.GetScheme()` to set the scheme in the `SetupUserDeactivationWebhooksWithManager()` function.

internal/controllers/iam/user_controller.go
JoseSzycho: Done! Thanks

scotwells (approved):

PR #227, opened 2025-08-13
feat(notification): Add EmailTemplate CRD for managing email templates

JoseSzycho:
## Description

This PR introduces a new EmailTemplate CRD under the `notification.miloapis.com` API group to support reusable email templates with variable substitution. This will enable standardized email communications across the platform, starting with user invitations.

### Changes

- Added new `notification.miloapis.com` API group
- Implemented EmailTemplate CRD with:
  - Support for HTML and plain text email bodies
  - Go template-based variable substitution
  - Variable type validation (string, url)
  - Required field validation
  - Cluster-scoped resource
- Added sample email template for user invitations

### DIFF
--- a/pkg/apis/notification/v1alpha1/emailtemplate_types.go
+++ b/pkg/apis/notification/v1alpha1/emailtemplate_types.go
@@ -57,6 +57,7 @@ type EmailTemplateSpec struct {
 
 	// Variables enumerates all variables that can be referenced inside the template expressions.
 	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:MaxItems=100
 	Variables []TemplateVariable `json:"variables,omitempty"`
 }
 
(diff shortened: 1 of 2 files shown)


pkg/apis/notification/v1alpha1/emailtemplate_types.go
```diff
@@ -0,0 +1,102 @@
...
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
+
+	// Required indicates whether the variable must be provided when rendering the template.
+	// +kubebuilder:validation:Required
+	Required bool `json:"required"`
+
+	// Type provides a hint about the expected value of this variable (e.g. plain string or URL).
+	// +kubebuilder:validation:Required
+	Type EmailTemplateVariableType `json:"type"`
```
scotwells, line 40: How would this be used internally? What if this needs to be a number?

JoseSzycho: I think that numbers should be treated as strings. Just to avoid confusion if some decimal needs to be sent. 
I wanted to make a point on the URL type, as I want to verify that the URL type variables are https and well formed

scotwells (approved):

PR #228, opened 2025-08-13
chore: get 100% on the go score card

scotwells:
A+ is nice, 100% is better. 

https://goreportcard.com/report/go.miloapis.com/milo


JoseSzycho:
@scotwells should we look for a GitHub actions that automatically do a check agains goreportcard?

scotwells:
@JoseSzycho pretty sure it just runs the go linters

PR #239, opened 2025-08-25
chore: add delete permission for IAM user self-management

JoseSzycho:
Updated `iam-user-self-manage` so users can delete their own account, as this is a needed feature:
https://github.com/datum-cloud/enhancements/issues/50


JoseSzycho:
@yahyafakhroji FYI

PR #277, opened 2025-09-16
Allow sending emails to external (non-Milo) addresses

JoseSzycho:
This change adds support for specifying a plain email address in notification.v1alpha1.Email, enabling Milo to dispatch emails to recipients who are not represented as Milo users or resources.


scotwells (dismissed):
Nice work on the CEL validation 💪

JoseSzycho:
@scotwells requesting review again, the image name got a different name after creating the PR

PR #305, opened 2025-10-01
feat: add validation package for quota resources

scotwells:
Introduces a new validation package that can be used to validate quota API resources. This includes a new resource type validator that uses an informer to maintain a cache of valid resource type configurations that can be used in resource claims.

### DIFF
--- a/internal/quota/validation/resourceclaim.go
+++ b/internal/quota/validation/resourceclaim.go
@@ -1,0 +1,105 @@
+package validation
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"k8s.io/client-go/dynamic"
+
+	quotav1alpha1 "go.miloapis.com/milo/pkg/apis/quota/v1alpha1"
+)
+
+// ValidationEngine provides validation capabilities for quota resources.
+type ValidationEngine interface {
+	// ValidateResourceClaimAgainstRegistrations validates that a ResourceClaim's
+	// ResourceRef is allowed to claim each of the requested resource types.
+	ValidateResourceClaimAgainstRegistrations(ctx context.Context, claim *quotav1alpha1.ResourceClaim) error
+
+	// ValidateClaimingResourcesConfiguration validates that all ClaimingResources
+	// in a ResourceRegistration are valid and that there are no duplicates.
+	ValidateClaimingResourcesConfiguration(registration *quotav1alpha1.ResourceRegistration) error
+}
+
+// validationEngine implements ValidationEngine using dynamic client for resource access
+// and ResourceTypeValidator for fast cached resource type validation.
+type validationEngine struct {
+	dynamicClient         dynamic.Interface
+	resourceTypeValidator ResourceTypeValidator
+}
+
+// NewValidationEngine creates a new validation engine with dynamic client support and
+// a required ResourceTypeValidator for optimized resource type validation.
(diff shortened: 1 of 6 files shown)


internal/quota/validation/resourcetype.go
```diff
@@ -0,0 +1,328 @@
...
+
+	// Cache for fast lookups
+	cacheMutex sync.RWMutex
+	cache      map[string]*claimingPermissions // resourceType -> claiming permissions
+}
+
+// NewResourceTypeValidator creates a new ResourceTypeValidator with async initialization.
+// This is the recommended approach as it prevents blocking during startup.
+// The validator will return immediately and sync in the background.
+func NewResourceTypeValidator(dynamicClient dynamic.Interface) ResourceTypeValidator {
```
joshlreese: Should this influence application readiness somehow? What's the issue with blocking on startup?

internal/quota/validation/resourcetype.go
```diff
@@ -0,0 +1,328 @@
...
+// ValidateResourceType validates using the cached data.
+// The cache only contains active ResourceRegistrations, so this is a simple lookup.
+func (v *resourceTypeValidator) ValidateResourceType(ctx context.Context, resourceType string) error {
+	v.cacheMutex.RLock()
+	defer v.cacheMutex.RUnlock()
+
+	_, exists := v.cache[resourceType]
+	if !exists {
+		//lint:ignore ST1005 "Error message intentionally capitalized for user-facing display"
+		return fmt.Errorf("Resource type '%s' is not available for quota management. Enable quota tracking for this resource type by registering it with the quota system", resourceType)
```
joshlreese: Should this instead return a structured error? Maybe [errors.NewInvalid](https://github.com/kubernetes/kubernetes/blob/9d5cda7a21fa7c1a9cde99e78fad468562b89ca4/staging/src/k8s.io/apimachinery/pkg/api/errors/errors.go#L284)?

internal/quota/validation/resourcetype.go
joshlreese: Nice!

internal/quota/validation/resourcetype.go
scotwells: I ended up leaving this alone because the function doesn't have a GVK to build a new invalid error. I'm leaving it to the callers to build a structured error.

internal/quota/validation/resourceclaim.go
```diff
@@ -0,0 +1,74 @@
...
+func NewResourceClaimValidator(dynamicClient dynamic.Interface, resourceTypeValidator ResourceTypeValidator) ResourceClaimValidator {
+	return &resourceClaimValidator{
+		dynamicClient:         dynamicClient,
+		resourceTypeValidator: resourceTypeValidator,
+	}
+}
+
+// ValidateResourceClaimAgainstRegistrations validates that the ResourceClaim's
+// ResourceRef is allowed to claim each of the requested resource types.
+func (v *resourceClaimValidator) ValidateResourceClaimAgainstRegistrations(ctx context.Context, claim *quotav1alpha1.ResourceClaim) error {
```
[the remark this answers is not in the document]
scotwells: Adjusted this to use a `field.ErrorList` in https://github.com/datum-cloud/milo/pull/340. I also renamed this method signature to just `Validate(ctx, claim)` to align with other validators.

PR #310, opened 2025-10-01
feat: add admission plugin for request-time quota enforcement

scotwells:
### Summary

Introduces an admission plugin that enforces quota limits at resource creation time by automatically creating and evaluating **ResourceClaims** based on **ClaimCreationPolicies**. Provides real-time quota enforcement with configurable timeout and retry behavior.

### How It Works

The admission plugin intercepts resource creation requests in the API server admission chain and:

1. **Policy Matching**: Finds ClaimCreationPolicies matching the incoming resource (GVK + CEL conditions), filtering to only `Ready=True` policies
2. **Claim Creation**: Creates a ResourceClaim on behalf of the user with labels:
   - `quota.miloapis.com/auto-created=true`
   - Annotation: `quota.miloapis.com/created-by=claim-creation-plugin`
3. **Decision Waiting**: Blocks the admission request while AllowanceBucketController (#309) evaluates quota and sets `Granted` condition
4. **Enforcement**: Allows resource creation if granted, rejects if denied or timeout exceeded

#### Admission Chain Position

The plugin runs in the API server admission chain:

```
HTTP Request
  ↓
Authentication
  ↓
Authorization
  ↓
MutatingAdmissionWebhook
  ↓
ClaimCreationPlugin  ← THIS PLUGIN (blocks until quota decision)
  ↓
ValidatingAdmissionWebhook
  ↓
Storage (etcd write)
```

**Why this position?**
- **After authentication**: Needs user info for CEL evaluation and claim ownership
- **Before storage**: Quota check must happen before resource is persisted
- **Before ValidatingWebhook**: Quota failures should happen early in the chain

#### Policy Evaluation

The plugin uses the PolicyEngine (#307) to:

- Find ClaimCreationPolicies matching the resource's GroupVersionKind
- Filter to only Ready=True policies
- Evaluate CEL trigger conditions with admission context (resource, user, request)
- Render ResourceClaim specs from Go templates
- Create claims with auto-created labels for lifecycle management

#### Wait Strategy

The plugin watches the created ResourceClaim for status updates:

- Uses Kubernetes watch API (event-driven, not polling)
- Responds immediately when AllowanceBucketController updates claim status
- Times out if no decision within configured duration (default: 30s)

#### Shared Informer Architecture

The plugin uses an efficient shared informer pattern for watching ResourceClaims:

- **Single watch connection**: One API server connection shared by all concurrent admission requests
- **Informer cache**: Reduces API server load by caching ResourceClaim state
[...]

PR #311, opened 2025-10-01
Integrate quota system into apiserver / controller manager

scotwells:


joshlreese (approved):
Approving, though I suspect there'll be a followup related to https://github.com/datum-cloud/milo/pull/310#discussion_r2434054494

PR #338, opened 2025-10-15
feat: add argo events to test environment w/ demo

scotwells:
Deploys argo events to Milo's test environment for event-driven automation based on Milo's API resources. Includes an example on how to use argo events to receive events from Milo's API and send a message to slack.

Closes: https://github.com/datum-cloud/enhancements/issues/402

### DIFF
--- a/Taskfile.yaml
+++ b/Taskfile.yaml
@@ -14,7 +14,7 @@ vars:
   MILO_IMAGE_TAG: "dev"
   TEST_INFRA_CLUSTER_NAME: "test-infra"
   # Test infra repository configuration - can be overridden with environment variable
-  TEST_INFRA_REPO_REF: 'v0.4.1'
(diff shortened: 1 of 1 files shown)


config/dependencies/argo-system/argo-events-helm-release.yaml
```diff
@@ -0,0 +1,54 @@
...
+    remediation:
+      retries: 30
+  upgrade:
+    remediation:
+      retries: 30
+  values:
+    # Deploy in namespaced mode - only watch resources in milo-system namespace
+    # This creates appropriate RBAC (Roles/RoleBindings instead of ClusterRoles)
+    # and limits the controller's scope to a single namespace
+    singleNamespace: true
```
zachsmith1: will this change in the future? just curious if there is anything outside of the milo namespace we would want to watch

PR #348, opened 2025-10-20
Enforce namespace consistency for ContactGroupMembership & MembershipRemoval when Contact targets resourcemanager API

JoseSzycho:
This PR introduces improved namespace validation logic for both ContactGroupMembership and ContactGroupMembershipRemoval resources. Specifically, it ensures that when a Contact uses a SubjectRef with APIGroup set to resourcemanager.miloapis.com, any associated membership or membership removal resource must be created in the same namespace as the referenced Contact.

This change enforces stricter scoping rules, improves organizational consistency, and prevents accidental misplacement or orphaning of membership and membership removal resources.


JoseSzycho:
@mustela FYI.

PR #309, opened 2025-10-01
feat: add core quota controllers

scotwells:
### Summary

This PR introduces the four fundamental controllers that form the operational core of the quota system. These controllers manage the lifecycle and validation of **[ResourceRegistration](https://github.com/datum-cloud/milo/blob/feature/quota-api-surface/docs/api/quota.md#resourceregistration)**, **[ResourceGrant](https://github.com/datum-cloud/milo/blob/feature/quota-api-surface/docs/api/quota.md#resourcegrant)**, **[ResourceClaim](https://github.com/datum-cloud/milo/blob/feature/quota-api-surface/docs/api/quota.md#resourceclaim)**, and **[AllowanceBucket](https://github.com/datum-cloud/milo/blob/feature/quota-api-surface/docs/api/quota.md#allowancebucket)** resources, implementing the actual quota accounting, aggregation, and enforcement logic.

### Core Controllers

**AllowanceBucketController** (Single writer pattern)
- Aggregates limits from ResourceGrants, tracks claim consumption
- Makes grant/deny decisions, creates buckets on-demand
- Owns allocations field via Server-Side Apply

**ResourceClaimController**
- Aggregates allocation status into Granted condition
- Read-only for quota decisions (defers to AllowanceBucket)
- Owns conditions field via Server-Side Apply

**ResourceGrantController**
- Validates resource types against active registrations
- Watches registrations for revalidation triggers
- Sets Active condition based on validation

**ResourceRegistrationController**
- Sets Active condition after CRD validation
- Minimal logic (validation via OpenAPI/CEL)

#### Architecture Highlights

- Single-writer pattern: Only AllowanceBucket modifies bucket status
- Server-Side Apply: Conflict-free updates with explicit field ownership
- Shared components: Single ResourceTypeValidator and CEL engine instance
- Initialization order: Registration → Grant → Bucket/Claim

#### Controller Setup

Unified initialization with shared validator, CEL engine, and InformerManager in dependency order.

#### Depends On:

#283 - Quota API definitions (ResourceRegistration, ResourceGrant, ResourceClaim, AllowanceBucket)
#305 - Validation package for ResourceType validation
#306 - Policy controllers (referenced in setup function)
#308 - Lifecycle controllers (referenced in setup function)

#### Required By:

#310 - Admission plugin (uses AllowanceBucketController's grant/deny decisions)
#311 - System integration (registers these controllers with controller manager)

### Testing

End-to-end testing for these controllers is provided in #315, including:

[...]

PR #351, opened 2025-10-21
feat: support multi-cluster quota enforcement

scotwells:
### Summary

This PR updates the quota system controllers to support managing quota resources across multiple control planes to support project-level resource quota enforcement. The Milo multi-cluster runtime provider is used to discover projects to connect to and manage quota resources.

This follows up with #346 to handle updating all of the quota controllers to support multi-cluster quota enforcement.

### Details

- All quota controllers updated to use the multi-cluster runtime library w/ the mil provider for project discovery. 
- The `milo-system` namespace is automatically created in a project when it's storage layer is initialized
- Allowance buckets are now created in the `organization-{org}` namespace when the consumer is an organization and the `milo-system` namespace when the consumer is a project.
- A new `multi-cluster-enforcement` end-to-end test was created to confirm that we can create project-level grants using a GrantCreationPolicy and enforce quota limits on `secrets` in the project. 
- Removed user 1001 from test bootstrapping since this is not necessary for the tests to execute.

### Notes

Follow up to #309 and #346

### DIFF
--- a/internal/controllers/resourcemanager/project_controller.go
+++ b/internal/controllers/resourcemanager/project_controller.go
@@ -149,13 +149,6 @@ func (r *ProjectController) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{RequeueAfter: 2 * time.Second}, nil
 	}
 
-	// Ensure per-project "milo-system" Namespace exists for quota system resources
-	if err := ensureMiloSystemNamespace(ctx, projCfg); err != nil {
-		logger.Error(err, "ensure milo-system namespace failed", "project", project.Name)
-		// Backoff and retry; don't mark Ready yet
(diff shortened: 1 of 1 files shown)


internal/controllers/resourcemanager/project_controller.go
```diff
@@ -248,6 +255,35 @@ func ensureDefaultNamespace(ctx context.Context, cfg *rest.Config) error {
 	return nil
 }
 
+// ensureMiloSystemNamespace ensures the milo-system namespace exists in the project control plane.
+// This namespace is used by the quota system to store ResourceGrants and AllowanceBuckets.
+func ensureMiloSystemNamespace(ctx context.Context, cfg *rest.Config) error {
```
joshlreese: Does this already get done in `bootstrapMiloSystemNamespace` in the project mux?

PR #363, opened 2025-10-23
Branding alignment

felixwidjaja:


felixwidjaja:
You're right. Thanks for the quick change.
I just pushed image update and dark/light mode for that. #366

Issue #361, opened 2025-10-23
UserInvitation creation fails when using Role with namespace datum-cloud


yahyafakhroji:
### **High-Level Summary**

Creating a new **UserInvitation** that references a Role within the `datum-cloud` namespace fails due to a validation error from the admission webhook, even though the Role exists and is accessible.
When the same Role reference is used under the `milo-system` namespace, the request succeeds (tested via `dryRun`).

---

### **Steps to Reproduce**

1. Send a POST request to create a `UserInvitation` using the `datum-cloud` namespace:

   ```bash
   curl -X POST \
     -H "Accept:application/json, text/plain, */*" \
     -H "Content-Type:application/json" \
     -H "Authorization:Bearer <access_token>" \
     --data '{
       "apiVersion": "iam.miloapis.com/v1alpha1",
       "kind": "UserInvitation",
       "metadata": { "name": "hiyahya-testing-cm715p-ounipaq9" },
       "spec": {
         "familyName": "Fakhroji",
         "givenName": "Yahya",
         "email": "molla.otoke@gmail.com",
         "expirationDate": "2025-10-24T07:53:28+07:00",
         "organizationRef": { "name": "hiyahya-testing-cm715p" },
         "roles": [
           { "name": "iam-editor", "namespace": "datum-cloud" }
         ],
         "state": "Pending"
       }
     }' \
     "https://api.staging.env.datum.net/apis/resourcemanager.miloapis.com/v1alpha1/organizations/hiyahya-testing-cm715p/control-plane//apis/iam.miloapis.com/v1alpha1/namespaces/organization-hiyahya-testing-cm715p/userinvitations?dryRun=All"
   ```
2. Observe the webhook rejection message:

   ```json
   {
     "status": "Failure",
     "reason": "Invalid",
     "message": "admission webhook \"vuserinvitation.iam.miloapis.com\" denied the request: UserInvitation.iam.miloapis.com \"hiyahya-testing-cm715p-ounipaq9\" is invalid: spec.roles[0].namespace: Invalid value: \"datum-cloud\": namespace is required"
   }
   ```
3. Modify the same payload but change `namespace` to `"milo-system"`.
4. The request succeeds.

---

### **Expected Behavior**

The `UserInvitation` should be valid when referencing a Role under the same organization namespace (`datum-cloud`), as long as the Role is accessible and exists in that namespace.

---

### **Actual Behavior**

The admission webhook rejects the request with the error:

```
[...]

PR #375, opened 2025-10-27
feat: Add viewer role and enhance owner role

JoseSzycho:
This commit introduces a new IAM role named `viewer`, providing view access to all resources with appropriate annotations for clarity. Additionally, the existing `owner` role is updated to include new annotations and several inherited roles, enhancing its capabilities and documentation.

Related to: https://github.com/datum-cloud/milo/issues/361#issuecomment-3451202248


JoseSzycho:
@scotwells Okay, so it is going to be a new different role

scotwells:
@JoseSzycho correct, it should be an Owner role in the `datum-cloud` namespace so it comes back in the list of roles available to organization admins.

PR #414, opened 2025-11-05
feat: add selectable field for user email in CRD

JoseSzycho:
This update introduces a new selectable field for the user email in the IAM user CRD, allowing for enhanced querying capabilities. The change is reflected in both the CRD definition and the corresponding API types.

This change was requested by @mustela in order to query Users by email.


JoseSzycho:
@scotwells is this safe to merge? There is a quota E2E test error

PR #426, opened 2025-11-08
feat: implement organization deletion on last member removal

JoseSzycho:
This update introduces functionality to automatically delete a Standard Organization when its last remaining OrganizationMembership references a deleted User.

The last remaining OrganizationMembership is used as a referenced as it cannot be deleted because of webhooks constraints.

### DIFF
--- a/config/controller-manager/overlays/core-control-plane/rbac/role.yaml
+++ b/config/controller-manager/overlays/core-control-plane/rbac/role.yaml
@@ -129,6 +129,15 @@ rules:
   - iam.miloapis.com
   resources:
   - platforminvitations
+  - users
+  verbs:
+  - get
+  - list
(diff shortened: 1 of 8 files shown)


drewr:
This behavior is counterintuitive to me. What's the link to the original discussion?

JoseSzycho:
@drewr Here's the link were we realized that we’re missing a clean up on standard organizations when the only user that’s a member of the organization is deleted https://datum-inc.slack.com/archives/C084W6XRVS7/p1762546366522879?thread_ts=1762543292.084779&cid=C084W6XRVS7

And here's the discussion were we discovered this issue: 
https://datum-inc.slack.com/archives/C084W6XRVS7/p1762543292084779

PR #433, opened 2025-11-17
feat: add UserSlackNotificationController for user creation notifications

JoseSzycho:
This commit introduces the UserSlackNotificationController, which sends a Slack notification whenever a new User is created. The controller checks for a Slack webhook URL and handles notifications accordingly. 

Screenshot 2025-11-17 at 5 53 36 PM

### DIFF
--- a/internal/controllers/iam/userslacknotification_controller.go
+++ b/internal/controllers/iam/userslacknotification_controller.go
@@ -1,0 +1,195 @@
+package iam
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"time"
+
+	iamv1alpha1 "go.miloapis.com/milo/pkg/apis/iam/v1alpha1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/retry"
+	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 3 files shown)


internal/controllers/iam/userslacknotification_controller.go
```diff
@@ -0,0 +1,195 @@
...
+		// Fallback to the email if no name components are present.
+		displayName = user.Spec.Email
+	}
+
+	// Example: "12 Jan 2025 14:32 UTC"
+	createdAt := user.CreationTimestamp.Time.UTC().Format("02 Jan 2006 15:04 MST")
+
+	createdText := fmt.Sprintf("*Created At:* %s", createdAt)
+
+	link := fmt.Sprintf("https://staff.datum.net/customers/users/%s", user.Name)
```
cc-datum: It would be helpful to load this from configuration to ensure the feature is environment agnostic.

internal/controllers/iam/userslacknotification_controller.go
```diff
@@ -0,0 +1,195 @@
...
+	createdText = fmt.Sprintf("%s\n\n<%s|View in Staff Portal>", createdText, link)
+
+	payload := map[string]interface{}{
```
cc-datum: A quick suggestion, using a template for this might be useful, especially if the structure becomes more complex later.

cc-datum (approved):
This LGTM 🌮 

I left minor comments that doesn’t necessarily need to be addressed in this PR, since this code may be removed once ArgoCD is fully rolled out

PR #436, opened 2025-11-19
Add docs for Marketing Contact migration into Milo

JoseSzycho:
This PR adds concise documentation for the Marketing Contact migration into the Milo cluster, including the steps, inputs, and example manifests used.

It exists so future migrations are reproducible, auditable, and easier to reason about when onboarding new teams or running similar data imports.

Related to: https://datum-inc.slack.com/archives/C084W6XRVS7/p1763480341060909

PR #438, opened 2025-11-19
feat: implement duplicate email validation across namespaces for Contacts

JoseSzycho:
This PR enhances the contact validation logic by adding checks to prevent the creation and update of contacts with duplicate email addresses across different namespaces. 

It introduces a new test case to verify that an error is returned when attempting to create a contact with an email that already exists in the cluster. Additionally, it updates the indexing mechanism to support email-based queries, ensuring data integrity in contact management.

During Contact Manual migration, we noticed that we were not preventing contacts with duplicated email addresses across all namespaces, which will cause inconsistencies on the email providers, as the email providers use email addresses as indexes.

Related to:
https://github.com/datum-cloud/milo/pull/436

### DIFF
--- a/test/notification/contactgroup-visibility/02-test-contact.yaml
+++ b/test/notification/contactgroup-visibility/02-test-contact.yaml
@@ -10,4 +10,4 @@ spec:
     name: admin
   familyName: Visibility
   givenName: Test
-  email: admin@test.local
+  email: visibility-test@test.local


scotwells (approved):

JoseSzycho:
Fixing the tests, seems that we are using duplicated email address on some contacts and tests stop working

PR #463, opened 2025-12-23
feat: add proxy client certificate support

scotwells:
## Summary

Add `--proxy-client-cert-file` and `--proxy-client-key-file` flags to enable mTLS authentication when proxying requests to aggregated API servers.

This allows aggregated API servers to trust authentication headers (`X-Remote-User`, `X-Remote-Group`, etc.) forwarded by Milo, resolving `system:anonymous` user issues.

## Detail

As I was working through deploying the activity-apiserver as an aggregated API on Milo's platform, I ran into issues where the activity apiserver was receiving requests with the `system:anonymous` user instead of the end-user I was authenticating as. 

I was able to trace this down to being caused by Milo not having a proxy client certificate issued that it uses to communicate with aggregated apiservers. 

---

Causes https://github.com/datum-cloud/enhancements/issues/536#issuecomment-3687933423

### DIFF
--- a/cmd/milo/apiserver/server.go
+++ b/cmd/milo/apiserver/server.go
@@ -140,7 +140,23 @@ func NewCommand() *cobra.Command {
 	s.Metrics.AddFlags(namedFlagSets.FlagSet("metrics"))
 	logsapi.AddFlags(s.Logs, namedFlagSets.FlagSet("logs"))
 	s.Traces.AddFlags(namedFlagSets.FlagSet("traces"))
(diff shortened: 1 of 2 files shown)


ecv (approved):

zachsmith1 (approved):
Does this affect the identity sessions aggregated api server? I thought we ran into this issue back when that work was done

scotwells:
@zachsmith1 no, that's configured separately.

PR #467, opened 2025-12-30
feat: implement api filter so users can only list public contact groups and related private contact groups

### LINKED ISSUE #466: Implement API Filter so users can only list Public Contact Groups
Currently, when users retrieve contact groups, they can list all of them, which is incorrect.

Private groups should only be retrieved if they have a ContactGroupMembership for that specific group.

JoseSzycho:
This PR implements a visibility filter for ContactGroup resources that restricts which ContactGroups users can see when listing through their control-plane endpoint. Public groups are visible to everyone, while private groups are only visible to users who have an associated ContactGroupMembership.

Relates to
- https://github.com/datum-cloud/enhancements/issues/511

Closes #466

### DIFF
--- a/pkg/server/filters/audit_scope_test.go
+++ b/pkg/server/filters/audit_scope_test.go
@@ -58,7 +58,7 @@ func TestDetermineScopeAnnotations(t *testing.T) {
 		{
 			name: "global scope (no parent)",
 			userInfo: &user.DefaultInfo{
-				Name: "test-user",
+				Name:  "test-user",
 				Extra: map[string][]string{},
 			},
 			want: map[string]string{


pkg/server/filters/contactgroups.go
```diff
@@ -0,0 +1,333 @@
...
+			captureWriter.flush()
+			return
+		}
+
+		// Filter the contact groups based on visibility
+		filteredBody, err := f.filterContactGroups(ctx, userID, captureWriter.body)
+		if err != nil {
+			responsewriters.InternalError(w, req, fmt.Errorf("failed to filter contact groups: %w", err))
+			return
+		}
```
scotwells: @JoseSzycho could we inject a field selector here to force the caller to only retrieve Public Contact Groups? Seems like that's what is essentially happening in the `f.filterContactGroups` function.

PR #475, opened 2026-01-08
chore: add selectable fields for userRef.name and emailAddress

JoseSzycho:
This PR allows Emails to be filtered by spec.recipient.userRef.name  or spec.recipient.userRef.emailAddress.

Related to:
- https://github.com/datum-cloud/staff-portal/issues/286

### DIFF
--- a/docs/api/identity.md
+++ b/docs/api/identity.md
@@ -1,4 +1,70 @@
 # API Reference
 
-Packages:
+## Packages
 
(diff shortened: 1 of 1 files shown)


docs/api/identity.md
scotwells: Any idea why these documentation changes were removed?

scotwells (approved):

PR #486, opened 2026-01-20
fix: quota allowance buckets not created

scotwells:
This PR resolves an [issue with pre-created allowance buckets not being created in projects](https://github.com/orgs/datum-cloud/discussions/156#discussioncomment-15542294). This resolves the issue by creating the allowance buckets as part of the resource grant controller which has correct access to the cluster context to connect to the correct control plane. 

The buckets were originally created in the request handler that was responsible for ensuing requests for the allowance bucket to reconcile. The functions in this context don't have access to the cluster the original request was being made from to create the bucket in the correct control plane.

I also adjusted the grant controller to only update the resource status when something changes to prevent unnecessary updates.


drewr (approved):
🙏 

> The functions in this context don't have access to the cluster the original request was being made from to create the bucket in the correct control plane.

So does that mean there was probably an access error happening somewhere? And if so was it just getting buried in the noise?

PR #488, opened 2026-01-21
feat: graduate Notes API to dedicated notes.miloapis.com group

OscarLlamas6:
- Create new notes.miloapis.com API group with Note (namespaced) and ClusterNote (cluster-scoped)
- Implement controllers with PolicyBinding support for both resource types
- Add validation and mutation webhooks for Note and ClusterNote
- Generate CRDs for notes.miloapis.com/v1alpha1
- Configure ProtectedResources for IAM integration
- Add domain note examples demonstrating all use cases
- Remove Notes from crm.miloapis.com (graduated to dedicated group)
- Clean up CRM API group structure for future CRM-specific resources
- Update all kustomizations and configurations

This allows Notes to be attached to any resource (including networking.datumapis.com/Domain) while keeping CRM API group available for future CRM-specific resources like Leads, Opportunities, and Accounts.

Needed for https://github.com/datum-cloud/enhancements/issues/501

### DIFF
--- a/cmd/milo/apiserver/config.go
+++ b/cmd/milo/apiserver/config.go
@@ -402,7 +402,7 @@ func DefaultBuildHandlerChain(apiHandler http.Handler, c *server.Config, loopbac
 	handler = datumfilters.UserContactListConstraintDecorator(handler)
 	handler = datumfilters.UserContactGroupMembershipListConstraintDecorator(handler)
 	handler = datumfilters.UserContactGroupMembershipRemovalListConstraintDecorator(handler)
-	handler = datumfilters.ContactGroupVisibilityDecorator(loopbackConfig)(handler)
+	handler = datumfilters.ContactGroupVisibilityWithoutPrivateDecorator(handler)
 	handler = genericapifilters.WithRequestInfo(handler, c.RequestInfoResolver)
 	handler = genericapifilters.WithRequestReceivedTimestamp(handler)
 	// handler = genericapifilters.WithMuxAndDiscoveryComplete(handler, c.lifecycleSignals.MuxAndDiscoveryComplete.Signaled())
--- a/config/crd/bases/notification/notification.miloapis.com_contactgroups.yaml
+++ b/config/crd/bases/notification/notification.miloapis.com_contactgroups.yaml
@@ -216,6 +216,8 @@ spec:
                 type: array
             type: object
         type: object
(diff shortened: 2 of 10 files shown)


internal/controllers/notes/clusternote_controller.go
```diff
@@ -0,0 +1,184 @@
...
+
+	if !policyBindingReady {
+		log.Info("PolicyBinding not ready, will retry", "message", policyBindingMessage)
+		return ctrl.Result{}, fmt.Errorf("waiting for PolicyBinding to become ready: %s", policyBindingMessage)
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (r *ClusterNoteController) ensureCreatorEditorPolicyBinding(ctx context.Context, clusterNote *notesv1alpha1.ClusterNote, creator *iamv1alpha1.User) (bool, string, error) {
```
JoseSzycho: Both controllers (clusternote and note) share most of the logic. While it’s beneficial to have separate controllers, I would suggest sharing all of the functions that would be common to both.

internal/webhooks/notes/v1alpha1/clusternote_webhook.go
```diff
@@ -0,0 +1,122 @@
...
+	if err := m.Client.Get(ctx, client.ObjectKey{Name: string(req.UserInfo.UID)}, creatorUser); err != nil {
+		return errors.NewInternalError(fmt.Errorf("failed to get user '%s' from iam.miloapis.com API: %w", string(req.UserInfo.UID), err))
+	}
+
+	clusterNote.Spec.CreatorRef = iamv1alpha1.UserReference{
+		Name: creatorUser.Name,
+	}
+
+	return nil
+}
```
JoseSzycho: Now that we’ve separated CRDs for cluster and namespaced resources, we should use the Defaulter to set the resource owner. This way, the notes will be selected for garbage collection when the owner is deleted.

JoseSzycho (changes requested):
Overall, great work.

Just left some comments about sharing functionalities and setting up the resources owners.

OscarLlamas6:
> Overall, great work.
> 
> Just left some comments about sharing functionalities and setting up the resources owners.
 
 Thanks for the feedback @JoseSzycho  ! I've implemented both suggestions. Let me know your feedback again please

PR #524, opened 2026-03-18
fix: handle stale ResourceClaims blocking quota admission retries

zachsmith1:
## Summary
- Fixed marshaling of unstructured object so that metadata was correctly usable within CEL
- Added `milo_quota_existing_claim_resolutions_total` counter (labels: `not_found`, `existing_granted`, `stale_deleted`, `delete_failed`, `get_failed`) to track how often each resolution path activates in production


scotwells (approved):
@zachsmith1 having a chainsaw end-to-end tests for this would give us a bit more confidence that this works as expected.

PR #523, opened 2026-03-18
feat: add quota registrations and claim policies for notes and core resources

zachsmith1:
## Summary

- Add ResourceRegistration and ClaimCreationPolicy definitions for 4 milo-native resources
- Follows the same pattern as NSO quotas (datum-cloud/network-services-operator#129)
- Grant policies with default amounts will be added separately in the datum repo

### Resources covered

| Resource | apiGroup | Consumer | Trigger apiVersion |
|----------|----------|----------|-------------------|
| Note | `notes.miloapis.com` | Project | `notes.miloapis.com/v1alpha1` |
| ClusterNote | `notes.miloapis.com` | Project | `notes.miloapis.com/v1alpha1` |
| Secret | `core.miloapis.com` | Project | `v1` |
| ConfigMap | `core.miloapis.com` | Project | `v1` |

## Test plan

- [ ] Verify kustomize build succeeds: `kustomize build config/quota/`
- [ ] Verify registrations are picked up by Milo quota system after infra wiring (separate PR)

Ref: datum-cloud/enhancements#664

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #528, opened 2026-03-19
fix: wire quota registrations and claim policies into services kustomization

zachsmith1:
## Summary

The `config/quota/` directory containing ResourceRegistrations and ClaimCreationPolicies for Notes, ClusterNotes, Secrets, and ConfigMaps was not referenced by any deployment kustomization. These resources were merged via PR #523 but would never be deployed by Flux.

Add `../quota` as a component to `config/services/kustomization.yaml` so the registrations and claim policies are deployed alongside other service configurations.

## Test plan

- [x] `kustomize build config/services/` succeeds and includes all ResourceRegistrations and ClaimCreationPolicies


config/services/kustomization.yaml
```diff
@@ -6,3 +6,4 @@ kind: Component
 
 components:
   - quota
+  - ../quota
```
scotwells: There's two quota folders?

PR #531, opened 2026-03-20
fix: skip quota enforcement for resources outside project context

zachsmith1:
## Summary

Two fixes for quota enforcement breaking existing e2e tests after ClaimCreationPolicies were deployed in PR #528:

### 1. Skip quota in root control plane

The admission plugin now skips ResourceClaim creation when no consumer (project) context can be determined. Resources created in the root control plane (e.g. CRM notes test) are not subject to per-project quota.

Previously, creating Notes, ConfigMaps, or Secrets outside a project control plane failed with "Insufficient quota" because the claim had no consumerRef and couldn't match any AllowanceBucket.

### 2. Add test-specific grants for project-scoped tests

Add ResourceGrant setup steps to `note-multicluster-subject` and `clusternote-multicluster-subject` e2e tests so their projects have quota allocated. These tests create resources in project control planes which now require quota grants.

## Test plan

- [x] Unit tests pass
- [ ] `crm-note-contact-lifecycle` passes (root-scoped Notes skip quota)
- [ ] `note-multicluster-subject` passes (project-scoped with test grant)
- [ ] `clusternote-multicluster-subject` passes (project-scoped with test grant)

### DIFF
--- a/test/notes/clusternote-multicluster-subject/chainsaw-test.yaml
+++ b/test/notes/clusternote-multicluster-subject/chainsaw-test.yaml
@@ -53,6 +53,23 @@ spec:
                 name: Ready
                 value: 'true'
 
+    - name: setup-quota-grant
+      description: Provision quota for the project so resources can be created
+      cluster: project
+      try:
+        - apply:
+            file: test-data/quota-grant.yaml
+        - wait:
+            apiVersion: quota.miloapis.com/v1alpha1
+            kind: ResourceGrant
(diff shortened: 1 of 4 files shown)


scotwells:
@zachsmith1 the diff here isn't matching the PR description and the tests are still failing. Is this actually ready for review?

zachsmith1:
@scotwells this actually might not be needed

PR #534, opened 2026-03-23
feat: add user agent field and updatedAt to session endpoint

mattdjenkinson:
This PR adds fields for location, client details and last active to the user session. Once this is merged i'll raise a PR for auth-provider-zitadel to populate these fields.


pkg/apis/identity/v1alpha1/types.go
```diff
@@ -12,13 +12,40 @@ type Session struct {
...
-	CreatedAt     metav1.Time  `json:"createdAt"`
-	ExpiresAt     *metav1.Time `json:"expiresAt,omitempty"`
+	// UserUID is the unique identifier of the user who owns this session.
+	UserUID string `json:"userUID"`
+
+	// Provider is the authentication provider for this session (e.g. "zitadel").
+	Provider string `json:"provider"`
+
+	// IP is the client IP address associated with the session, if known.
+	IP string `json:"ip,omitempty"`
```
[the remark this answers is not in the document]
scotwells: @mattdjenkinson have you confirmed that's how Zitadel sessions works? I would expect a refresh token to be valid regardless of IPs changing for the session. IPs can change frequently these days, so that seems like a high friction point.

pkg/apis/identity/v1alpha1/types.go
```diff
@@ -12,13 +12,40 @@ type Session struct {
...
+
+	// FingerprintID is an optional device or client fingerprint from the provider.
+	FingerprintID string `json:"fingerprintID,omitempty"`
+
+	// CreatedAt is when the session was created.
+	CreatedAt metav1.Time `json:"createdAt"`
+
+	// ExpiresAt is when the session expires, if applicable.
+	ExpiresAt *metav1.Time `json:"expiresAt,omitempty"`
```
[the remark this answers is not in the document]
mattdjenkinson: Zitadel computes this from the `Login Lifetimes` we set in the console, so it would be 10 days. That's the maximum time a session can be active for.

kevwilliams (dismissed):

scotwells (dismissed):
Looks good. Let's test with Zitadel before we merge this in. Assume you confirmed all these fields are available from Zitadel?

JoseSzycho:
@mattdjenkinson seems that when you executed `task generate` , some  manual generated API docs got deleted, as the `identity API`

JoseSzycho (dismissed):

config/controller-manager/overlays/core-control-plane/rbac/role.yaml
scotwells, line 1: What caused these to get removed?

mattdjenkinson: Not deliberate — these are an auto-generated drift fix from `task generate`.

The committed `role.yaml` on `main` was out of sync with the kubebuilder RBAC markers in source. The only markers for `userinvitations` live at `internal/controllers/iam/userinvitation_controller.go:92-93`:

```go
// +kubebuilder:rbac:groups=iam.miloapis.com,resources=userinvitations,verbs=get;list;watch;update;delete
// +kubebuilder:rbac:groups=iam.miloapis.com,resources=userinvitations/status,verbs=update
```

Neither marker grants `patch` on `userinvitations`, nor mentions `userinvitations/finalizers` at all. So when I re-ran `task generate` post-merge, controller-gen produced the output that matches today's markers — and the stale lines fell out.

If the controller actually needs `patch` and/or finalizer access, the fix is to extend the markers (e.g. `verbs=get;list;watch;update;patch;delete` and add a `userinvitations/finalizers` marker) and regenerate. Happy to do that here if it was an unintentional drop in source. Otherwise the regen is the right state.

scotwells (dismissed):

config/controller-manager/overlays/core-control-plane/rbac/role.yaml
mattdjenkinson: Followed up in 1de1096 — added the missing markers at `internal/controllers/iam/userinvitation_controller.go:92-94` so controller-gen produces the same RBAC main was carrying and the source-of-truth lives next to the controller code:

```go
// +kubebuilder:rbac:groups=iam.miloapis.com,resources=userinvitations,verbs=get;list;watch;update;patch;delete
// +kubebuilder:rbac:groups=iam.miloapis.com,resources=userinvitations/status,verbs=update
// +kubebuilder:rbac:groups=iam.miloapis.com,resources=userinvitations/finalizers,verbs=update
```

The regenerated role.yaml now matches main's pre-merge state for these two rules.

scotwells (approved):

PR #547, opened 2026-03-28
Add grace period to project stuck creating alert

scotwells:
## Summary

Adds a 2-minute grace period to the ProjectStuckCreatingSLOViolation alert rule. Previously, the alert fired immediately (`for: 0s`), which meant any brief gap in monitoring data would trigger a false page.

## Context

On March 28, a short monitoring database disruption caused this alert to fire incorrectly. The 2-minute grace period gives the system time to recover from brief outages before paging on-call.

Tracking issue: https://github.com/datum-cloud/infra/issues/2059


ecv (approved):
Can we log the outages instead of simply not paging on them?

PR #553, opened 2026-04-01
Fix project creation delays caused by slow deletions

scotwells:
> [!NOTE]
> CI shows 6 failing e2e tests — these are pre-existing failures unrelated to this PR. See #549 for the fix.

## Summary

- Project creation could stall for up to 10 minutes while another project was being deleted. The controller now handles deletions without blocking other work.
- Deletion progress is visible via a new `ResourceCleanup` status condition on the project, so operators can see exactly what phase the cleanup is in.
- The controller now processes up to 4 projects concurrently instead of 1.

## What changed

The project controller previously ran a synchronous cleanup operation during deletion that could block for up to 10 minutes. During that time, no other project — including newly created ones — could be reconciled.

The cleanup is now split into two non-blocking steps: issuing delete commands (fast) and checking whether resources have drained (cheap poll). The controller tracks progress via a `ResourceCleanup` condition with three reasons:
- `CleanupStarted` → delete commands are being issued
- `CleanupAwaitingCompletion` → waiting for resources to be removed
- `CleanupComplete` → all resources removed, finalizer can be released

If resources haven't drained after a check, the controller transitions back to `CleanupStarted` to automatically re-issue delete commands on the next reconcile.

## Test plan

- [x] Existing unit tests pass (`task test:unit`)
- [ ] Delete a project and verify the `ResourceCleanup` condition progresses through `CleanupStarted` → `CleanupAwaitingCompletion` → `CleanupComplete`
- [ ] Create a new project while another project is being deleted — creation should not be delayed
- [ ] Kill the controller pod mid-deletion and verify cleanup resumes after restart
- [x] End-to-end tests pass (`task test:end-to-end`) — new `project-deletion` test passes in 20.7s

### DIFF
--- a/internal/controllers/projectpurge/purge.go
+++ b/internal/controllers/projectpurge/purge.go
@@ -2,8 +2,11 @@ package projectpurge
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"net"
 	"strings"
+	"syscall"
 	"time"
 
 	"golang.org/x/sync/errgroup"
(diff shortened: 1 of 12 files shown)


ecv (approved):
that joggr thing is kinda cool.

PR #560, opened 2026-04-02
fix: eliminate 502 errors during rolling deployments

### LINKED ISSUE #559: Reduce API errors during rolling deployments
## Problem

When the Milo API server is updated, clients occasionally receive brief 502 errors. This happens because the gateway continues sending requests to pods that are shutting down before it realizes they're no longer available.

The issue resolves itself within minutes, but it creates noisy alerts and a degraded experience for anyone making API calls during that window.

## Expected Behavior

Deployments should be seamless — clients should never see errors when the API server is being updated.

## Proposed Solution

Add an Envoy Gateway `BackendTrafficPolicy` to the existing `gateway-api` kustomize component alongside the HTTPRoute. This gives operators a safe operational baseline out of the box without needing to configure it themselves.

The policy should include:

- **Retries** on gateway errors so failed requests are transparently re-sent to a healthy pod
- **Passive health checking** to quickly eject unhealthy pods from the load balancer pool
- **Active health checking** so the gateway detects pod readiness faster than the default Kubernetes endpoint propagation

Additionally, add a `PodDisruptionBudget` to ensure a minimum number of API server pods remain available during voluntary disruptions (node drains, cluster upgrades).

## Context

This is a recurring issue observed in staging. Most recently on 2026-04-02 where a single 502 was returned to a client calling the OpenAPI discovery endpoint during a routine image update rollout.

scotwells:
## Summary

Clients occasionally see brief 502 errors when the API server is updated. This PR makes deployments seamless so clients never see errors during rollouts.

**What changed:**
- The gateway now automatically retries failed requests on a healthy pod instead of returning the error to the client
- Health checks detect pods shutting down within seconds and stop sending them traffic
- The API server signals it's going away before actually stopping, giving the gateway time to react
- A disruption budget prevents too many pods from going down at once during cluster maintenance
- The rollout strategy now waits for a new pod to be fully ready before terminating the old one

Closes #559

## Test plan

- [ ] Deploy to staging with 2+ API server replicas
- [ ] Trigger a rolling update (image tag change) while sending continuous API requests
- [ ] Verify zero 502 errors are returned to clients during the rollout
- [ ] Verify the `BackendTrafficPolicy` is accepted by Envoy Gateway (`kubectl get btp -n milo-system`)
- [ ] Test a voluntary node drain to confirm the PDB keeps at least one pod available

### DIFF
--- a/config/apiserver/deployment.yaml
+++ b/config/apiserver/deployment.yaml
@@ -10,7 +10,7 @@ spec:
       app.kubernetes.io/part-of: milo-control-plane
   strategy:
     rollingUpdate:
-      maxSurge: 1
+      maxSurge: 25%
       maxUnavailable: 0
     type: RollingUpdate
   template:
--- a/config/apiserver/deployment.yaml
+++ b/config/apiserver/deployment.yaml
@@ -10,8 +10,8 @@ spec:
       app.kubernetes.io/part-of: milo-control-plane
   strategy:
     rollingUpdate:
-      maxSurge: 25%
-      maxUnavailable: 25%
+      maxSurge: 1
+      maxUnavailable: 0
     type: RollingUpdate
   template:
     metadata:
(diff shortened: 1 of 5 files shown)


JoseSzycho (dismissed):

config/apiserver/deployment.yaml
```diff
@@ -11,7 +11,7 @@ spec:
   strategy:
     rollingUpdate:
       maxSurge: 25%
-      maxUnavailable: 25%
+      maxUnavailable: 0
```
ecv, line 14: hahaha

ecv (dismissed):

ecv (approved):

ecv:
@scotwells sorry this slipped me, let's merge it?

PR #563, opened 2026-04-13
feat: add ActivityPolicy for MachineAccount resources

kevwilliams:
## Summary

- Adds `config/milo/activity/policies/machineaccount-policy.yaml` with an `ActivityPolicy` for `iam.miloapis.com/MachineAccount`
- Covers create, delete, deactivate, activate, and generic update audit rules
- State-specific rules (`deactivate`/`activate`) match on `spec.state` to produce meaningful summaries for activation state changes
- No event rules — the controller sets conditions but does not emit Kubernetes Events
- Follows the same `config/milo/activity/` convention established in the dns-operator repo

## Test plan

- [ ] Apply to staging with `kubectl apply -k config/milo/activity/` (or wire into the Flux Kustomization)
- [ ] Verify policy is `Ready=True` via `datumctl get activitypolicies`
- [ ] Create a MachineAccount and confirm an activity entry appears
- [ ] Update `spec.state` to `Inactive` and confirm "deactivated" activity appears
- [ ] Update `spec.state` back to `Active` and confirm "reactivated" activity appears
- [ ] Delete a MachineAccount and confirm a delete activity appears


kevwilliams:
Tried working out how to add these policies for MachineAccounts, if its in the right direction I can continue with some others but thought this would be as good as any place to start. @scotwells

config/services/identity/policies/kustomization.yaml
scotwells, line 1: These files should go under `config/milo/identity/policies`.

scotwells (commented):
Looks directionally accurate. Have you tried this against the policy preview endpoint to confirm it gives you the timeline you expect?

config/services/identity/policies/kustomization.yaml
kevwilliams: Ran it through the PolicyPreview endpoint — all rules fired correctly:
  - create → kevin@datum.net created machine account ci-bot
  - delete → kevin@datum.net deleted machine account old-bot
  - deactivate → kevin@datum.net deactivated machine account ci-bot
  - activate → kevin@datum.net reactivated machine account ci-bot
  - System controller status patch → no match (correctly filtered)
 
Also moved the files per your other comment

scotwells: Oh sorry, I gave the wrong path. It should have been in `config/services/identity/policies` to line up with https://github.com/datum-cloud/milo/blob/main/config/services/kustomization.yaml.

kevwilliams:
Moved again in 3de08dc — now at `config/services/identity/policies/` and wired into `config/services/kustomization.yaml`.

scotwells (dismissed):

scotwells:
@kevwilliams any reason you removed it from the service's kustomization?

kevwilliams:
Re-added in 19185ad. Removed it because a CI run failed with `no matches for kind "ActivityPolicy"` — assumed the activity CRDs weren't in the test cluster. The `clusternote-multicluster-subject` failure in the latest run is pre-existing on main and unrelated to this PR.

scotwells:
@kevwilliams oh, they probably aren't installed in the test cluster. We should probably re-evaluate what that test is doing because it's odd it's including the services configuration.

kevwilliams:
Yea so I removed it because the test-environment-validation CI job failed with:
  no matches for kind "ActivityPolicy" in version "activity.miloapis.com/v1alpha1"

It looks like the test environment spins up a fresh cluster and applies config/services/ directly. Since the activity service (which provides the ActivityPolicy CRD) isn't installed in that cluster, it couldn't apply the policy and the deploy step failed.

scotwells:
@kevwilliams can you disable that test for now? It's been broken for awhile and need to be fixed.

kevwilliams:
Got this to pass, also created https://github.com/datum-cloud/milo/issues/565 to deal with an issue separately, skipped that test for now.

scotwells (approved):

PR #566, opened 2026-04-15
Unify machine account resources under the identity API

scotwells:
## What changes

Machine account records and their credentials now live together under one API group (`identity.miloapis.com`). Milo hands off storage of the machine account itself to the auth provider, which already owns the credentials.

## Why

A machine account and its keys are two halves of the same thing. Keeping them in different API groups made the surface confusing and forced clients to talk to two places for one conceptual resource.

## What to expect

- The old `iam.miloapis.com/machineaccounts` path stops resolving. Clients should use `identity.miloapis.com/machineaccounts`.
- Machine account keys are now cluster-scoped (no namespace).
- Any RBAC rules naming the old group need updating.

Paired with datum-cloud/auth-provider-zitadel#102 — the two must deploy together.

## Test plan

- [x] Build, vet, and tests pass
- [ ] Validate in staging alongside the paired PR


scotwells:
Closing unstarted. The auth-provider-zitadel companion PR (datum-cloud/auth-provider-zitadel#102) depended on adding etcd-backed MachineAccount storage in an aggregated apiserver, which would have required reimplementing Milo's project-scoped storage tenancy. Rather than duplicate that primitive, holding off on this change until we decide whether to export Milo's projectstorage wrapper or revisit the overall approach. Branch preserved for reference.

PR #571, opened 2026-04-19
Add metrics for all Milo resources and conditions

scotwells:
Enables metrics collection for every resource across all core Milo services — IAM, identity, infrastructure, resource manager, and quota.

Each resource now exposes:
- Identity and creation time
- All status conditions automatically, so new condition types appear in dashboards without any config changes

Metrics are organized per service rather than bundled together, making it easier to deploy and manage coverage for each service independently.

## Test plan

- [ ] Deploy to a dev environment and confirm metrics appear in Victoria Metrics for IAM, identity, infrastructure, resource manager, and quota resources
- [ ] Verify that all condition types (not just `Ready`) show up as separate series per resource


drewr:
I'm trying to understand the net change here. We weren't reporting on things like creation time and status conditions before?

scotwells:
@drewr creation time / status conditions were available in the control plane. This configures the [resource-metrics service](https://github.com/milo-os/resource-metrics) to surface this data as time series data so we can understand how this data changes over time. Only the quota resources had their resource info surfaced as telemetry data. 

We're moving to the resource-metrics service because a single deployment can collect metrics across all projects instead of deploying an instance of kube-state-metrics per project like we are today.

drewr:
Got it. This is enabling the pivot from kube-state-metrics. Thanks!

JoseSzycho (approved):

Issue #512, opened 2026-03-04
bug: insufficient quota resources - staging


mattdjenkinson:
In staging I sometimes run into a `Insufficient quota resources available. Review your quota usage and reach out to support if you need additional resources.` error when creating DNS record sets. My project quota says i've only used 6/500.

<img width="884" height="472" alt="Image" src="https://github.com/user-attachments/assets/7610fe0e-62d7-4743-ad65-e6b794b2f811" />
<img width="1292" height="420" alt="Image" src="https://github.com/user-attachments/assets/1d2cdc47-dd3f-40b1-a323-ba91d6bf85be" />

Milo logs

```
"ts": 1772643311147.774,   "caller": "admission/plugin.go:302",   "msg": "ResourceQuotaEnforcement admission plugin triggered",   "v": 3,   "logger": "resource-quota-enforcement-plugin",   "project": "personal-project-86a07d72",   "operation": "CREATE",   "resource.group": "dns.networking.miloapis.com",   "resource.version": "v1alpha1",   "resource.kind": "DNSRecordSet",   "name": "mdj-test-online-wyt655-alias",   "namespace": "default",   "user": "mjenkinson@datum.net",   "dryRun": false 

```

```
"ts": 1772643414749.242,
  "caller": "admission/plugin.go:502",
  "msg": "ResourceClaim not granted, denying resource creation",
  "logger": "resource-quota-enforcement-plugin",
  "policy": "project-dnsrecordset-claim-policy",
  "resourceName": "test-mdj-test-online-ww0r9p-81b11f6f",
  "gvk": "dns.networking.miloapis.com/v1alpha1, Kind=DNSRecordSet",
  "err": "failed to create ResourceClaim: failed to create ResourceClaim: resourceclaims.quota.miloapis.com \"dnsrecordset-test-mdj-test-online-ww0r9p-81b11f6f\" already exists"
```

mattdjenkinson:
Update 27.04.26: have a PR open here that should resolve this. https://github.com/datum-cloud/milo/pull/573

mattdjenkinson:
In prod as of https://github.com/milo-os/milo/releases/tag/v0.25.1

PR #573, opened 2026-04-23
fix: surface precise admission errors and clear stale ResourceClaims

mattdjenkinson:
## The bug

Ref: https://github.com/datum-cloud/milo/issues/512

Users occasionally saw

> `Error from server (Forbidden): ... Insufficient quota resources available. Review your quota usage and reach out to support if you need additional resources.`

when creating resources such as DNS record sets, even though their project quota showed only a handful of resources in use (for example 6/500). The message was misleading because the admission plugin returned it for *every* failure inside the ResourceClaim pipeline, not just actual denials.

## Root cause

The `ResourceQuotaEnforcementPlugin` has three distinct failure paths, all of which were collapsed into the same 403 body:

1. **Actual denial** — `AllowanceBucketController` marks `Granted=False` with reason `QuotaExceeded`.
2. **Watch timeout** — the admission watch manager times out after 30 s because the bucket controller is still reconciling, or grants are still propagating.
3. **Stale `ResourceClaim`** — claim names are deterministic, so a leftover claim from a prior admission timeout blocks every retry with `AlreadyExists`. `DeniedAutoClaimCleanupController` only reaped `Denied` claims, not pending/timed-out ones, so these orphans could linger indefinitely.

Case 3 produced the "insufficient quota" message even when the user had plenty of quota available, which is exactly what the reporter hit.

### 1. Typed claim-failure kinds with distinct user-facing messages
`internal/quota/admission/errors.go` (new) introduces `claimFailure` with four kinds: `Denied`, `Timeout`, `Conflict`, `Internal`. `createAndWaitForResourceClaim` now returns these and `processResourceWithPolicy` maps each to a warm, actionable 403 body:

| Kind | Message |
|---|---|
| Denied | _You've reached your quota for this resource type (...). Delete unused resources to free up capacity, or contact support to request a higher limit._ |
| Timeout | _Your request took too long to be checked against your quota. Please try again in a moment — if this keeps happening, contact support._ |
| Conflict | _We're still cleaning up from a previous attempt to create this resource (...). Please try again in a few seconds._ |
| Internal | _Something went wrong while checking your quota for this request. Please try again — if this keeps happening, contact support._ |

### 2. Pre-create handling of existing claims
Before `Create`, `createResourceClaim` now does a `Get` at the deterministic name and branches:

| Existing claim state | Action |
|---|---|
| `Granted=True` | Short-circuit admission to allow (capacity already reserved for this resource) |
| `Granted=False` / `QuotaExceeded` | Return `Conflict` so the user gets a retry message, not "insufficient quota" |
| Pending or no condition | Delete inline, then Create a fresh claim |
| `AlreadyExists` on Create | Also mapped to `Conflict` |

Only auto-created claims (label `quota.miloapis.com/auto-created=true`) are touched, so hand-authored claims sharing a name are never disturbed.

### 3. Garbage collection for stale pending auto-created claims
`DeniedAutoClaimCleanupController` now also deletes auto-created claims that remain in a non-final state for longer than `DefaultStalePendingClaimAge` (5 min, well above the 30 s admission watch timeout). Fresh pending claims are requeued for the remaining time rather than deleted, so in-flight admissions are never disturbed. Denied-claim behavior is unchanged.

### Observability
New counter `milo_quota_admission_existing_claim_resolution_total{resolution=...}` tracks how often each pre-create path fires (`granted`, `denied`, `deleted_pending`, `create_conflict`), so the stale-claim behavior is no longer invisible.

## Files touched

- `internal/quota/admission/errors.go` (new)
- `internal/quota/admission/plugin.go`
- `internal/quota/admission/plugin_test.go`
- `internal/quota/controllers/lifecycle/cleanup.go`
- `internal/quota/controllers/lifecycle/cleanup_test.go` (new)

## Test plan

- [x] \`go test ./internal/quota/...\` passes locally
- [x] \`TestResolveExistingClaim\` covers granted short-circuit, denied conflict, pending delete+recreate, and clean-slate create
- [x] \`TestUserFacingClaimError\` asserts each claim-failure kind produces its expected message
- [x] \`TestClaimWaitScenarios\` extended with a \`timeout\` case (plus the mock now matches \`watchManager.evaluateClaimStatus\` semantics)
- [x] \`TestDeniedAutoClaimCleanupController\` covers denied-deleted, stale-pending-deleted, fresh-pending-requeued, no-condition-pending-deleted, granted-ignored, and manual-ignored
- [x] Manual verification: create a DNS record set, kill the admission plugin mid-flight to leave a pending claim, then retry and confirm the retry succeeds rather than returning "insufficient quota"
- [x] Check the new \`milo_quota_admission_existing_claim_resolution_total\` counter on a staging cluster

Made with [Cursor](https://cursor.com)


zachsmith1:
good catch. iirc if something happens inside the admission the rejection looks like its a quota issue even though its not related

internal/quota/admission/plugin.go
```diff
@@ -738,12 +850,50 @@ func (p *ResourceQuotaEnforcementPlugin) createResourceClaim(ctx context.Context
...
+				p.logger.Info("Failed to delete stale pending ResourceClaim before retry",
+					"claimName", claimName,
+					"namespace", namespace,
+					"error", delErr)
+				// Treat as conflict so the user retries once GC catches up.
+				return newConflictFailure("unable to clean up stale pending claim", delErr)
+			}
+			p.logger.V(2).Info("Deleted stale pending ResourceClaim before retry",
+				"claimName", claimName,
+				"namespace", namespace)
```
scotwells, line 883: Do you have any concrete steps on how to reproduce this issue? The auto-deletion that's happening here should already happen through normal quota operations.

mattdjenkinson: You can re-create this by creating an AI Edge, putting an invalid value in the origin field and then submitting the form multiple times.

scotwells (approved):

PR #591, opened 2026-04-30
feat: notify Slack when a user is held for review by fraud score

mattdjenkinson:
## Summary
- Adds `UserFraudSlackNotificationController` that posts to the existing user-bot Slack channel when a `FraudEvaluation` (`fraud.miloapis.com/v1alpha1`) completes with `status.decision != ACCEPTED` (i.e. `REVIEW` or `DEACTIVATE`) — the cases where a new user does **not** get auto-approved because of their MaxMind / minFraud score.
- Watches `FraudEvaluation` via `unstructured.Unstructured`, so milo does not pick up a new Go module dependency on the fraud manager.
- Reuses `MILO_USER_SLACK_WEBHOOK_URL` (no new secret / env var); the controller is disabled when that env var is unset, mirroring `UserSlackNotificationController`.
- Idempotency guard: `iam.miloapis.com/fraud-risk-slack-notified=true` annotation on the `FraudEvaluation`, plus an old-vs-new predicate so history-append updates don't re-fire.
- Slack payload includes display name, email, decision, composite score, and the staff-portal link (`https://staff.datum.net/customers/users/{name}`); red attachment color to differentiate from the yellow signup ping.

## Test plan
- [ ] `go build ./...` and `go vet ./...` pass (verified locally).
- [ ] Deploy to a dev cluster with `MILO_USER_SLACK_WEBHOOK_URL` pointing at a sandbox webhook.
- [ ] Apply a `User` and a `FraudEvaluation` with `status.phase=Completed, status.decision=REVIEW, status.compositeScore=85` referencing it; confirm one Slack message lands and the annotation is set.
- [ ] Re-update the same `FraudEvaluation` (e.g. append a history entry); confirm no second message is sent.
- [ ] Apply a `FraudEvaluation` with `decision=ACCEPTED`; confirm no Slack message and no annotation.
- [ ] Unset `MILO_USER_SLACK_WEBHOOK_URL`; confirm both slack controllers log "disabled" and do not register.

### DIFF
--- a/cmd/milo/controller-manager/controllermanager.go
+++ b/cmd/milo/controller-manager/controllermanager.go
@@ -713,6 +713,15 @@ func Run(ctx context.Context, c *config.CompletedConfig, opts *Options) error {
 					logger.Error(err, "Error setting up user Slack notification controller")
 					klog.FlushAndExit(klog.ExitFlushTimeout, 1)
 				}
+
+				userFraudSlackCtrl := iamcontroller.UserFraudSlackNotificationController{
+					Client:          ctrl.GetClient(),
+					SlackWebhookURL: userSlackWebhookURL,
+				}
+				if err := userFraudSlackCtrl.SetupWithManager(ctrl); err != nil {
+					logger.Error(err, "Error setting up user fraud Slack notification controller")
+					klog.FlushAndExit(klog.ExitFlushTimeout, 1)
+				}
 			}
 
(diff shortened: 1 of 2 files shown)


scotwells:
@mattdjenkinson we could probably solve this with a prometheus alert?

scotwells:
@mattdjenkinson not always, you can configure routing rules with alert manager so you can control where / how alerts are surfaced.

PR #599, opened 2026-05-05
ci: speed up e2e build and cluster setup

kevwilliams:
## Summary

- Add BuildKit \`--mount=type=cache\` for \`/go/pkg/mod\` and \`/root/.cache/go-build\` in the Dockerfile so the Go compiler and module caches persist across runs via the GHA layer cache
- Move \`generate:code\` to its own step before the Docker build — it has no dependency on the cluster and was blocking behind KinD creation
- Parallelize \`test-infra:cluster-up\` and the Docker image build by starting cluster creation in the background immediately before the \`docker/build-push-action\` step

## Background

The \`test-environment-validation\` job was spending ~13 minutes before tests even started:

| Step | Time |
|------|------|
| Build Milo container image | ~6m41s |
| Set up test infrastructure cluster | ~5m27s |

**Build slowness:** \`cache-from: type=gha\` caches Docker layers, but \`COPY . .\` invalidates the layer on every PR, triggering a full \`go build\` from scratch. Without \`--mount=type=cache\`, the Go build cache is discarded between runs.

**Sequential bottleneck:** Cluster creation and the Docker build are fully independent — they only converge at \`dev:load\`. Running them one after the other wastes ~5 minutes of wall-clock time.

## How the parallelism works

Cluster creation is IO-bound (pulling images, waiting on pod readiness) while the Docker build is CPU-bound, so they coexist well on the same runner. A "Start cluster creation in background" step fires \`task test-infra:cluster-up\` into the background before the \`docker/build-push-action\` step runs. A "Wait for cluster creation" step afterward collects the exit code and surfaces the full log on failure.

## Expected improvement

| Change | Savings |
|--------|---------|
| Go build cache mounts | ~3-5m after first warm cache hit |
| Parallel cluster + image build | ~5m per run |

- **Cold run** (first after merge): similar to before
- **Warm runs**: should drop from ~13m to ~8m before tests start

## Test plan

- [ ] Confirm the workflow triggers on this PR
- [ ] Verify "Generate code" step passes and generated files are available for the Docker build
- [ ] Check cluster creation log surfaces correctly in the "Wait for cluster creation" step
- [ ] Confirm "Build Milo container image" and cluster creation visibly overlap in the step timestamps
- [ ] Check that the Docker build step is faster on a second run of the same branch (warm cache)
- [ ] Confirm e2e tests still pass end-to-end

### DIFF
--- a/.github/workflows/test-environment-validation.yaml
+++ b/.github/workflows/test-environment-validation.yaml
@@ -72,6 +72,12 @@ jobs:
       - name: Generate code
         run: task generate:code
 
+      - name: Start cluster creation in background
+        run: |
+          task test-infra:cluster-up > /tmp/cluster-up.log 2>&1 &
+          echo $! > /tmp/cluster-up.pid
+          echo "Cluster creation started in background (PID: $(cat /tmp/cluster-up.pid))"
+
       - name: Build Milo container image with caching
         uses: docker/build-push-action@v5
         with:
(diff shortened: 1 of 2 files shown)


ghost:
## 📝 Documentation Analysis

All docs are up to date! 🎉



---

✅ Latest commit analyzed: 668e2fc8b2541503b6047bf60f9344197266f92e | Powered by [Joggr](https://joggr.ai?utm_source=gh&utm_medium=gh&utm_campaign=ghc&utm_id=ghc&utm_content=voltagent%2Fjoggr)

kevwilliams:
Closing because of inactivity, if we need it, we can open something again.

PR #601, opened 2026-05-06
feat: add cloud-portal-usage-metering feature flag

### LINKED ISSUE #711: feat: implement feature flagging via Milo entitlements
# Feature Flags, Without Building a Feature-Flag System

Implements the design from #577 (specified in #695).

## What the user actually gets

A way to turn product features on and off **per organization** — gated rollouts, private betas, paid add-ons, kill switches — without standing up LaunchDarkly, without writing a new controller, and without adding a single CRD.

If you can already grant a customer a quota of 50 VMs, you can already grant them a feature. It's the same machinery.

## How it works

A feature flag is just three existing Milo objects wearing a different hat:

| Concept | Implementation |
|---|---|
| The flag itself | `ResourceRegistration` with `type=Feature` |
| "Org X has it turned on" | `ResourceGrant` with `amount=1` targeting that org |
| "Is it on right now?" | `AllowanceBucket.status.available > 0` |

### Example: shipping a private beta of "AI Edge DNS"

**1. Define the flag** (one-time, by the service owner):

```yaml
apiVersion: quota.miloapis.com/v1alpha1
kind: ResourceRegistration
metadata:
  name: feature-ai-edge-dns
spec:
  type: Feature
  claimingResources: []
```

**2. Turn it on for Acme Corp** — staff users get a new UI in the **staff portal** (under the quota service view) where they can browse registered features and toggle them on or off for any organization. Each toggle creates or deletes a \`ResourceGrant\` like this one under the hood:

```yaml
apiVersion: quota.miloapis.com/v1alpha1
kind: ResourceGrant
metadata:
  name: acme-ai-edge-dns
spec:
  resource: feature-ai-edge-dns
  amount: 1
  target: { kind: Organization, name: acme }
```

No deploy, no PR, no engineering ticket — sales or support flips the switch and it takes effect on the next check.

**3. Check it from code** — through the standard OpenFeature SDK:

```go
if client.BooleanValue(ctx, \"ai-edge-dns\", false, evalCtx) {
    return enableEdgeDNSPath(req)
}
return legacyDNSPath(req)
```

```ts
if (await client.getBooleanValue(\"ai-edge-dns\", false)) {
  showEdgeDNSPanel();
}
```

Under the hood the OpenFeature provider does a single field-selector query against \`AllowanceBucket\` for the caller's org. Revoke the grant in the staff portal and the next check returns \`false\` — instant kill switch.

## How this changes the world

- **No parallel control plane.** Today, \"does this customer get X?\" lives in code, in Slack DMs, in someone's head. After this, it lives in the same place quotas already live — auditable, RBAC-gated, GitOps-friendly.
- **Sales and ops can flip flags without a deploy.** Staff users open the staff portal, find the org, and toggle the feature. Granting a beta to a design partner stops being an engineering task.
- **Entitlements and quotas are the same primitive.** A \"Pro plan\" can be expressed as: \`feature-sso = 1\`, \`feature-audit-log = 1\`, \`vms = 100\`. No second system to reconcile.
- **Standard SDK, no lock-in.** Code calls OpenFeature, not a Datum-specific API. If we ever swap the backend, callers don't change.
- **Zero new infrastructure to operate.** v1 ships with no new CRDs, no new controllers, no new binaries — just a thin provider library, a \`type=Feature\` enum value, and a staff-portal UI on top of existing APIs.

The punchline: feature flagging stops being a thing we have to build and becomes a thing that already exists, because we already built quotas.

## Technical Goals

Ordered backend → frontend, ending in docs and support readiness. Each step is a verifiable gate.

- [x] **Schema.** Add `Feature` to `ResourceRegistration.spec.type` enum in Milo. No new CRDs, controllers, or binaries. 
- [x] **Config + RBAC.** Ship Kustomize config and the `feature-flag-operator` IAM role for toggling grants.
- [x] ~**Backend provider (Go).** OpenFeature provider backed by an `AllowanceBucket` informer. Steady-state check is in-memory; one field-selector `LIST` on cold start. Closed by default on miss/error.~ Doing later
- [x] **Frontend provider (TypeScript).** Same evaluation contract as the Go provider, used by cloud portal and staff portal.
- [x] **Feature Gate Used** Gate a specific feature in the cloud-portal
- [x] **Staff-portal UI.** Browse registered features, toggle on/off per org. Each toggle = one `ResourceGrant` create/delete. RBAC enforced server-side.
- [x] **Auditability.** Every toggle is a `ResourceGrant` create/delete in the k8s audit log, attributable to a staff user. No parallel audit path.
- [x] **Failure mode.** Flag-system degradation does not fail the data path; services fall back to caller-supplied defaults.
- [x] **Docs.** Two short guides: (a) service developer — registering a flag and gating code with OpenFeature; (b) operator — toggling flags in the staff portal and reading the audit trail.
- [x] **Support readiness.** Runbook covering "flag not taking effect," "wrong org granted," and "provider erroring" — with the exact `kubectl`/`datumctl` queries to diagnose each.

## Non-Goals (v1)

- No per-user, per-project, or percentage rollouts. Org-scoped boolean only.
- No non-boolean flag values.
- No A/B or experimentation framework.
- No customer-facing flag management.

kevwilliams:
## Summary

Adds the first feature flag `ResourceRegistration` to the Milo bundle, along with the `config/services/features/` Kustomize scaffold.

This flag controls visibility of the Usage & Metering tab in the Datum Cloud portal — when granted to an organization the tab is shown, otherwise hidden.

## Changes

- `config/services/features/` — new Kustomize component wired into `config/services/`
- `config/services/features/registrations/cloud-portal-usage-metering.yaml` — `ResourceRegistration` with `type=Feature`

## Notes

- `claimingResources` uses `features.miloapis.com/FeatureGrant` sentinel to satisfy `minItems=1` without enabling admission enforcement
- `ResourceGrant` instances (granting this flag to specific orgs) live in `datum-cloud/infra`, not here
- Operators granting this flag need the existing `quota.miloapis.com-manager` role — no new role needed

Closes datum-cloud/enhancements#711

### DIFF
--- a/config/services/features/kustomization.yaml
+++ b/config/services/features/kustomization.yaml
@@ -1,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1alpha1
+kind: Component
+
+components:
+  - registrations
--- a/config/services/features/registrations/cloud-portal-usage-metering.yaml
+++ b/config/services/features/registrations/cloud-portal-usage-metering.yaml
@@ -1,0 +1,21 @@
+apiVersion: quota.miloapis.com/v1alpha1
+kind: ResourceRegistration
+metadata:
+  name: feature-cloud-portal-usage-metering
+  labels:
+    app.kubernetes.io/name: milo
+    app.kubernetes.io/component: feature-flags
+    features.miloapis.com/feature-name: cloud-portal-usage-metering
+spec:
+  consumerType:
+    apiGroup: resourcemanager.miloapis.com
+    kind: Organization
+  type: Feature
+  resourceType: features.miloapis.com/cloud-portal-usage-metering
+  description: "Controls visibility of the Usage tab in the Datum Cloud portal. When enabled for an organization, the Usage & Metering section is shown in their portal navigation."
+  baseUnit: feature
+  displayUnit: features
+  unitConversionFactor: 1
+  claimingResources:
+    - apiGroup: features.miloapis.com
+      kind: FeatureGrant
--- a/config/services/features/registrations/kustomization.yaml
+++ b/config/services/features/registrations/kustomization.yaml
@@ -1,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1alpha1
+kind: Component
+resources:
+  - cloud-portal-usage-metering.yaml
--- a/config/services/kustomization.yaml
+++ b/config/services/kustomization.yaml
@@ -8,4 +8,8 @@ components:
   - iam
   - resourcemanager
   - quota
-  - activity
+  - features
+  # identity is intentionally excluded here — it contains ActivityPolicy resources
+  # which require the activity.miloapis.com CRDs to be installed. Those CRDs are
+  # provided by the activity service, which is not deployed in the milo test cluster.
+  # Deploy config/services/identity/ separately once the activity service is present.


scotwells:
@kevwilliams let's define this in the billing service since it's a feature flag for the billing component.

kevwilliams:
Moving this to `datum-cloud/billing` per @scotwells — feature flags for a specific service should be defined and owned by that service, not the platform bundle.

PR #579, opened 2026-04-27
feat: Add ActivityPolicy resources for all core API groups

JoseSzycho:
## Summary

- Adds ActivityPolicy resources for 23 resource kinds across 6 API groups (`iam`, `resourcemanager`, `identity`, `notes`, `notification`, `quota`), organized under `config/services/activity/policies/<apigroup>/`
- Replaces the old `config/services/identity/policies/` catch-all with per-API-group folders
- Fixes a bug in the GroupMembership policy where `memberRef` (non-existent field) was used instead of 
- The `config/services/kustomization.yaml` now includes the `activity` component alongside `quota`; ActivityPolicy resources require the activity service CRDs (see companion datum-infra PR)

## Test plan

- [ ] `datumctl apply -k . --platform-wide` applies all policies without errors
- [ ] `datumctl get activitypolicies --platform-wide` shows all 23 policies with `Ready=True`
- [ ] Create a Group/User/Organization and verify activity events appear in the timeline
- [ ] Verify GroupMembership create/delete events show correct user and group names

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

Closes #572

PR #605, opened 2026-05-06
fix(services): exclude activity from default aggregate

kevwilliams:
## Summary

Remove ` activity` from the default `config/services/kustomization.yaml` aggregate so its `ActivityPolicy` resources aren't applied to clusters that don't have the `activity.miloapis.com` CRDs installed.

## Why

PR #579 added 23 `ActivityPolicy` resources under `config/services/activity/policies/...` and wired `activity` into the top-level `config/services/kustomization.yaml`, but the `ActivityPolicy` CRD is not part of the milo control plane bootstrap. As a result, every PR's `test-environment-validation` job fails when `task dev:deploy` runs `kubectl apply -k config/services/`:

```
resource mapping not found for name: "..." namespace: "" from "config/services/":
  no matches for kind "ActivityPolicy" in version "activity.miloapis.com/v1alpha1"
ensure CRDs are installed first
```

`config/services/activity/kustomization.yaml` already explicitly documents that this component should be applied separately once the activity service is present:

```yaml
# ActivityPolicy resources require the activity.miloapis.com CRDs to be installed.
# Those CRDs are provided by the activity service, which is not deployed in the
# milo test cluster. Deploy this directory separately once the activity service is present.
```

But the aggregate kept pulling it in regardless. This change makes the comment match the behavior.

## Behavior change

- `kustomize build config/services/` no longer emits `ActivityPolicy` resources (verified: 0 emitted, was 23).
- `kustomize build config/services/activity/` still emits all 23 `ActivityPolicy` resources for environments that need them.
- Test-environment bootstrap stops failing on missing `ActivityPolicy` CRD.

## Test plan

- [x] `kustomize build config/services/ | grep -c ActivityPolicy` → `0`
- [x] `kustomize build config/services/activity/ | grep -c ActivityPolicy` → `23`
- [ ] `test-environment-validation` job passes on this PR

### DIFF
--- a/config/services/kustomization.yaml
+++ b/config/services/kustomization.yaml
@@ -3,9 +3,13 @@ kind: Component
 
 # This component aggregates all Milo service configurations.
 # Add new service directories here as they are created.
+#
+# `activity` is intentionally NOT included here: ActivityPolicy resources
+# require the activity.miloapis.com CRDs, which are not part of the milo
+# control plane. Deploy `config/services/activity/` explicitly in
+# environments where the activity service is present.
 
 components:
   - iam
   - resourcemanager
   - quota
-  - activity


ghost:
## 📝 Documentation Analysis

All docs are up to date! 🎉



---

✅ Latest commit analyzed: 9c00baa3d80726b242cb730d0bafb5d77900aa63 | Powered by [Joggr](https://joggr.ai?utm_source=gh&utm_medium=gh&utm_campaign=ghc&utm_id=ghc&utm_content=voltagent%2Fjoggr)

scotwells (approved):
You'll want a corresponding change on the infra side to enable this.

Issue #616, opened 2026-05-14
DiscoveryContextFilter drops built-in groups (coordination.k8s.io, etc.) from project-context discovery


mattdjenkinson:
## Summary

When the `DiscoveryContextFilter` feature gate is enabled, kubectl / controller-runtime clients hitting the milo apiserver through a project-routed URL (e.g. `/apis/resourcemanager.miloapis.com/v1alpha1/projects/X/control-plane/apis`) see an **empty list of API groups** for built-in groups like `coordination.k8s.io`. The underlying REST handlers still serve the resources correctly — direct GETs at the raw URL (`/apis/coordination.k8s.io/v1/namespaces/.../leases/...`) succeed — but discovery never advertises them, so RESTMapper-based clients fail with `no matches for kind "Lease" in version "coordination.k8s.io/v1"`.

By design, built-in groups should pass straight through the filter: `Registry.AllowedContexts({coordination.k8s.io, leases})` returns nil because there's no CRD annotation and no `RegisterStatic` call, so `Matches(nil, ContextProject)` returns true, so `IsVisible` is true. They should be kept.

## Impact

This breaks any external operator that watches Lease (or any built-in resource) in a per-project apiserver. In staging, this manifested as `network-services-operator`'s connector + iroh-dns controllers failing to engage on 221 project control planes — the `Lease` watch RESTMapping fails, multicluster-runtime rejects the entire controller/cluster pair, and `Connector.status.conditions[Ready]` froze indefinitely. The in-reconcile `cl.Get(&coordinationv1.Lease{}, …)` similarly errored on every reconcile attempt.

Workaround landed in NSO: datum-cloud/network-services-operator#161.

## Environment

- milo image: `ghcr.io/milo-os/milo:v0.0.0-main-20260511-233451` (essentially equivalent to `v0.25.0`)
- Feature gates: `EventsProxy=true,ServiceAccountKeys=true,DiscoveryContextFilter=true`
- Staging only (production has the gate disabled)

## Suspected root cause

Commit `6ef5237` ("fix: detect aggregated discovery from request Accept header") changed `filterAPIIndex` to decide whether to decode the response as `apidiscoveryv2.APIGroupDiscoveryList` based solely on the **request** Accept header:

```go
if isAggregatedDiscovery(req.Header.Get("Accept")) {
    var list apidiscoveryv2.APIGroupDiscoveryList
    if err := json.Unmarshal(cw.body.Bytes(), &list); err != nil {
[...]

PR #617, opened 2026-05-15
fix: allow coordination.k8s.io/leases in Project discovery + body-kind probe defensive

mattdjenkinson:
## Summary

Two related changes:

### 1. Policy fix (the actual bug that broke staging)

`config/discovery/core/discovery-context-policy.yaml` listed `coordination.k8s.io/leases` as Platform-only — alongside genuinely admin-only groups like `rbac.authorization.k8s.io`, `authentication.k8s.io`, `admissionregistration.k8s.io`. That's incorrect. Leases are a namespace-scoped primitive used legitimately by per-project workloads: leader election in any operator targeting project apiservers, or custom heartbeats like Datum Connect's connector health Lease.

Effect of the misconfiguration: project-context aggregated discovery omitted `coordination.k8s.io` entirely. Any controller-runtime client failed to find `Lease` in its REST mapper. Downstream symptom: [`datum-cloud/network-services-operator#160`](https://github.com/datum-cloud/network-services-operator/issues/160) — connector and iroh-dns controllers' Lease watches failed to engage, Connector status froze, and the operator burned an engage-retry storm against every affected project control plane (221 of them on staging).

Fix: add `"Project"` to the Lease contexts.

### 2. Body-kind-probe defensive fix in `filterAPIIndex`

While debugging the symptom I traced it to a plausible-but-wrong cause: `filterAPIIndex` used the request `Accept` header to decide whether to decode the response as `apidiscoveryv2.APIGroupDiscoveryList`. If the inner apiserver ever emitted legacy `APIGroupList` JSON in response to an aggregated request, `json.Unmarshal` would succeed with `items: nil` and the filter would silently emit an empty discovery list — every group disappears.

That turned out **not** to be the cause of the user-visible symptom (the policy was), but the silent-empty failure mode is still a real footgun. Probe the response body's `kind` field first; treat the request Accept header as a tie-break for bodies with no explicit kind; pass unknown shapes through verbatim.

Keeping this defensive change in the same PR so the filter is robust to format mismatches going forward — but the *user-visible* fix is the policy change.

## Why this regressed only on staging

Production runs `v0.25.0` with `DiscoveryContextFilter` feature gate **disabled**, so the policy is never consulted and discovery passes through unfiltered. Staging enabled the gate (infra commit `adfe2b19`, 2026-04-17) which loaded the policy and exposed the misconfiguration.

## Test plan

New `TestFilterAPIIndex_FormatDetection` (4 sub-cases) covers the body-kind probe branches: aggregated body + aggregated Accept → filtered; legacy body + aggregated Accept → pass-through (no silent emptying); legacy body + legacy Accept → pass-through; aggregated body + legacy Accept → still filtered (body kind wins).

All existing tests pass.

The policy change should be validated end-to-end against an affected staging project after this PR lands — confirmed via an in-cluster `curl` against `/apis/resourcemanager.miloapis.com/v1alpha1/projects/X/control-plane/apis` with the aggregated `Accept` header, asserting that `coordination.k8s.io` appears in the response items.

## Related

- Fixes `#616`
- Downstream: [`datum-cloud/network-services-operator#160`](https://github.com/datum-cloud/network-services-operator/issues/160) (frozen Connectors), [`datum-cloud/network-services-operator#161`](https://github.com/datum-cloud/network-services-operator/pull/161) (closed without merge; defensive code that's no longer needed once this fix lands)


pkg/server/discovery/filter.go
```diff
@@ -105,25 +115,32 @@ func filterAPIIndex(w http.ResponseWriter, req *http.Request, next http.Handler,
...
+		(kindProbe.Kind == "" && isAggregatedDiscovery(req.Header.Get("Accept"))):
 		var list apidiscoveryv2.APIGroupDiscoveryList
 		if err := json.Unmarshal(cw.body.Bytes(), &list); err != nil {
 			cw.flushUnchanged()
 			return
 		}
 		filterAggregatedGroups(&list, registry, parentCtx)
 		cw.flushJSON(&list)
-		return
+	case kindProbe.Kind == "APIGroupList":
```
savme: let's drop this branch since it's the same as `default`?

savme (dismissed):

savme (approved):

Issue #631, opened 2026-05-29
The controller-manager grows until it OOMs and evicts its node


ecv:
## Summary

`milo-controller-manager` (image `ghcr.io/milo-os/milo:v0.26.1`) exhausts its memory in production, hitting both its cgroup limit (cgroup OOMKill) and driving GKE nodes into memory-pressure eviction. The controller-manager replica is repeatedly OOMKilled/evicted, and co-scheduled pods (promtail, node_exporter) are evicted as collateral.

## Environment

- Image: `ghcr.io/milo-os/milo:v0.26.1`
- Deployment: `milo-controller-manager`, namespace `datum-system`, `replicas: 1`
- Args: `controller-manager`, `--control-plane-scope=core`, `--leader-elect=true`, `-v=6`
- Resources: `requests: cpu=500m, memory=5Gi` / `limits: cpu=2, memory=6Gi`
- Platform: GKE (datum infra-prod)

## Observed behavior

Live pod memory at time of diagnosis: **5.26 GiB working set** vs **6Gi limit** (~88%), trending upward.

On 2026-05-29 between ~20:27–21:20 UTC, across two nodes:

**Node `gke-...-2436176a-hdjd`**
- `EvictionThresholdMet` x7 ("Attempting to reclaim memory")
- `NodeHasInsufficientMemory` x6
- cgroup `OOMKilling`: `Memory cgroup out of memory: Killed process ... (milo) total-vm:7269832kB`
- milo-controller-manager pod evicted repeatedly (`The node was low on resource: memory. Threshold quantity: 100Mi`)

**Node `gke-...-63eaa943-5rsx`**
- Same `EvictionThresholdMet` / `NodeHasInsufficientMemory` pattern
- cgroup `OOMKilling`: `Killed process ... (milo) total-vm:7333136kB`
- Collateral evictions/OOMKills: `node_exporter`, `promtail-8c5rd`, `promtail-bjw4j`

Two distinct failure modes, same root cause:
1. **cgroup OOM** — milo process exceeds the 6Gi container limit and is killed by the kernel.
2. **node memory pressure** — kubelet evicts pods once node `available < 100Mi`; milo replica churn plus collateral eviction of neighboring pods.

## Expected behavior

`milo-controller-manager` memory usage stays bounded well under its limit over time.

## Impact

- Controller-manager enters an OOMKill/eviction loop, degrading reconciliation availability.
- Node memory pressure evicts unrelated workloads (logging/metrics agents) sharing the node.

[...]

PR #632, opened 2026-05-30
fix(controller-manager): scope per-project quota caches to cut memory

ecv:
## Problem

The Milo multicluster provider engages one full controller-runtime cluster per Ready Project (`pkg/multicluster-runtime/milo/provider.go:251`, `cluster.New`), and the only `ClusterOption` set is the `Scheme` (`cmd/milo/controller-manager/controllermanager.go:630-634`). Each per-project cache therefore stores **full objects, including `managedFields`**, for every type the provider-engaged quota controllers watch.

Memory scales linearly with project count. In production ~395 projects drive the single controller-manager replica past its 6Gi limit:

| Metric | Value |
|---|---|
| project control-plane namespaces | 395 |
| `go_goroutines` | 438,125 (~1,100 / project) |
| `go_memstats_stack_inuse_bytes` | 1.51 GiB |
| heap + stack | 4.87 GiB |
| `container_memory_working_set_bytes` | 5.51 GiB |

Result: cgroup OOMKills of the `milo` process and node memory-pressure evictions (incl. collateral promtail / node_exporter).

Full diagnosis: #631.

## Change

Scope each per-project cache to the three GVKs the provider-engaged quota controllers actually watch (`ResourceClaim`, `ResourceGrant`, `AllowanceBucket` — `internal/quota/controllers/core/{claim,grant,bucket}.go`, all `WithEngageWithProviderClusters(true)`), and strip `managedFields` from all cached objects via `cache.TransformStripManagedFields()`.

```go
o.Cache = cache.Options{
    ByObject: map[client.Object]cache.ByObject{
        &quotav1alpha1.ResourceClaim{}:   {},
        &quotav1alpha1.ResourceGrant{}:   {},
        &quotav1alpha1.AllowanceBucket{}: {},
    },
    DefaultTransform:            cache.TransformStripManagedFields(),
    ReaderFailOnMissingInformer: true,
}
```

## Expected impact

- `DefaultTransform: TransformStripManagedFields()` is the primary win — applies across all ~395 project caches, typically 30-50% per-object heap, cutting a large chunk of the ~3.35 GiB informer heap. No behavioral change: reconcilers do not read `managedFields`.
- Does **not** reduce the 1.51 GiB of goroutine stacks (one reflector set per project). That needs an architectural follow-up: shard projects across replicas, lazy/idle cache eviction, or a shared cache.

## Verification

- `go build ./cmd/milo/...` — passes
- `go vet ./cmd/milo/controller-manager/` — passes
- Symbols confirmed in vendored controller-runtime v0.23.3 (`pkg/cache/cache.go`: `ByObject`, `DefaultTransform`, `ReaderFailOnMissingInformer`, `TransformStripManagedFields`).

## Review note / risk

`ReaderFailOnMissingInformer: true` makes cached reads of any type **outside** the three above fail on project clusters (instead of silently creating a wide informer). Please confirm no provider-cluster controller does a cached `Get`/`List` of another type before merging. The ResourceClaimOwnershipController already uses a direct dynamic client (`internal/quota/controllers/lifecycle/ownership.go:226`), not the cache. If any such read exists, drop `ReaderFailOnMissingInformer` or add the type to `ByObject`.

## Recommended companion changes (separate)

- Set `GOMEMLIMIT` (~90% of the memory limit) on the deployment — converts residual spikes into GC backpressure instead of OOMKills.
- Apply the same `cache.Options` scoping to the core manager + infraCluster.
- Fix the `cancelFns` leak / add Ready→NotReady teardown in `provider.go`.

Draft pending the review note above.

### DIFF
--- a/cmd/milo/controller-manager/controllermanager.go
+++ b/cmd/milo/controller-manager/controllermanager.go
@@ -28,6 +28,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apiserver/pkg/server/healthz"
 	"k8s.io/apiserver/pkg/server/mux"
+	apiservercompat "k8s.io/apiserver/pkg/util/compatibility"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	cacheddiscovery "k8s.io/client-go/discovery/cached/memory"
 	"k8s.io/client-go/dynamic"
@@ -43,6 +44,7 @@ import (
 	certutil "k8s.io/client-go/util/cert"
 	cliflag "k8s.io/component-base/cli/flag"
 	"k8s.io/component-base/cli/globalflag"
+	basecompatibility "k8s.io/component-base/compatibility"
 	"k8s.io/component-base/configz"
 	"k8s.io/component-base/featuregate"
 	"k8s.io/component-base/logs"
@@ -53,8 +55,6 @@ import (
 	"k8s.io/component-base/term"
 	"k8s.io/component-base/version"
 	"k8s.io/component-base/version/verflag"
-	basecompatibility "k8s.io/component-base/compatibility"
-	apiservercompat "k8s.io/apiserver/pkg/util/compatibility"
 	genericcontrollermanager "k8s.io/controller-manager/app"
 	"k8s.io/controller-manager/controller"
 	"k8s.io/controller-manager/pkg/clientbuilder"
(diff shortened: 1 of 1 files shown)


ecv:
Can't believe it's this simple. But, spent 350K tokens across seven agents finding it. So, maybe it's not that simple. 

Relates to https://github.com/milo-os/inventory/issues/16

scotwells:
@savme could you take a look at this when you get a sec? Relates to the recent work you did around resource metrics.

savme (approved):
:shipit: :shipit: :shipit:

PR #639, opened 2026-06-01
docs(controller-manager): assess disabling embedded GC at core scope

ecv:
## Summary

Assessment (RFC) of whether the embedded Kubernetes garbage collector (GC) controller should be disabled when the controller-manager runs at `--control-plane-scope=core`, to cut goroutine/memory usage.

Doc only — no behavioral change. Follow-up to #631; complements #632 (which stripped managedFields from per-project quota caches).

## Document

`docs/proposals/core-scope-embedded-gc-assessment.md`

It covers:
- How the embedded GC is wired here (with `file:line` references).
- The monitor-per-GVK cost model (goroutines + metadata heap) and that it's measurable today via `--controllers=*,-garbagecollector`.
- Which cascade-deletion paths are covered without GC (namespace content deleter, project purger, explicit IAM finalizers) vs. which **rely** on the embedded GC (User-owned PolicyBinding/UserPreference; downstream anchor ConfigMaps; possibly membership PolicyBindings).
- Options: leave enabled / disable via existing flag / default-disable in code at core scope.
- Recommendation: do not disable yet; measure cost and prove cascade coverage first, with a concrete staging validation plan.

## Note

This is an assessment with open questions and a measurement plan, not a proposal to disable. It explicitly does not claim disabling is safe today.

Refs #631, #632


ecv:
Seeing if this is safe.

scotwells:
@ecv no, we should never disable garbage collection. It's used to clean up resources from the core control plane.

PR #640, opened 2026-06-01
docs(controller-manager): RFC for sharding projects across replicas

ecv:
## Summary

RFC / design proposal (no production code changes) for sharding Project cluster engagements across N controller-manager replicas.

The controller-manager engages one full controller-runtime `cluster.Cluster` per Ready Project (cache + informers + reflectors), and with leader election + `replicas: 1` a single active replica holds them all. This makes the goroutine count and stack memory scale linearly with project count: ~1,100 goroutines/project, 1.51 GiB `go_memstats_stack_inuse_bytes` at ~395 projects (#631). #632's managedFields strip cut per-object heap but does **not** reduce this goroutine-stack floor.

The RFC proposes distributing per-project engagements across replicas (active/active) so each pod holds ~projects/N caches → ~1/N of the stack floor per pod.

## Document

`docs/proposals/shard-projects-across-replicas.md`

Covers: the goroutine-stack floor problem, three sharding approaches (static label-selector / hash-based / lease-per-shard) with trade-offs and a recommendation (hash-based via StatefulSet ordinal, HRW hashing), leader-election → active/active migration and failover, a concrete change sketch in `provider.go`/`controllermanager.go`, a metrics-based rollout/validation plan, and open risks (rebalancing churn, lease scoping vs. singleton controllers, webhook traffic, and quota cross-cluster coordination correctness — flagged as highest risk).

This is an **RFC / design proposal only** — no implementation.

Refs #631, follow-up to #632.


scotwells:
@savme can you look into this when you have some spare cycles?

savme:
@scotwells oh, this looks great! Definitely happy to dig into it.

PR #643, opened 2026-06-02
fix(quota): stop ownership controller from garbage-collecting manually created ResourceClaims

### LINKED ISSUE #642: Ownership controller incorrectly garbage-collects manually-created ResourceClaims
## Problem

The `ResourceClaimOwnershipController` deletes any granted ResourceClaim that lacks `ownerReferences` after a 30-second orphan grace period. This was designed to clean up claims created by the admission plugin's `claim-creation-plugin` that never received an owner reference.

The problem is that the controller applies this logic to **all** ResourceClaims, not just the ones the admission plugin created. Services that manage their own ResourceClaim lifecycle (e.g. creating a claim per resource instance and tying cleanup to a finalizer) get caught in a continuous delete-recreate loop:

1. Service creates ResourceClaim (no `ownerReferences`)
2. Milo grants it — `Granted=True`
3. Ownership controller fires, can't resolve the owner (resource lives in a different control plane), returns 404
4. After 30s grace + 30s max-age → **DELETE**
5. Service sees 404, recreates → repeat

In production this produces ~600 create/delete cycles per day per project, `allocated=0` in the AllowanceBucket despite active resources consuming quota, and `QuotaGranted` flapping on every affected resource.

## Expected behaviour

The ownership controller should only manage claims that were auto-created by the admission plugin. Manually-created claims that carry their own lifecycle should be left untouched.

## Root cause

The ownership controller's watch predicate gates only on `Granted=True && len(ownerRefs)==0`. It has no equivalent of the check already present in `DeniedAutoClaimCleanupController`, which correctly limits itself to claims bearing:

```
Labels:      quota.miloapis.com/auto-created = "true"
Annotations: quota.miloapis.com/created-by  = "claim-creation-plugin"
```

Manually-created claims have neither marker, but the ownership controller processes them anyway.

scotwells:
## Problem

The `ResourceClaimOwnershipController` was incorrectly processing **any** `ResourceClaim` with `Granted=True` and no owner references — including claims created manually or by external services. Because it found no recognizable owner for these service-managed claims, it entered a continuous delete-recreate loop, causing:

- ~600 churn cycles per day in affected clusters
- `allocated=0` reported in `AllowanceBuckets`  
- `QuotaGranted` condition flapping

The `DeniedAutoClaimCleanupController` in the same package already handles this correctly by gating on the admission-plugin markers. The ownership controller simply lacked the same guard.

## Solution

- Extract `isAutoCreatedClaim` as a **package-level function** in `ownership.go`, making it the single source of truth for both controllers (the cleanup controller's method now delegates to it).
- Add the guard to the **watch predicate** in `SetupWithManager` so manually created claims are never enqueued.
- Add the guard to **`Reconcile`** as defence-in-depth, catching any claims already in the queue at rollout time.

## Test plan

- [x] `TestIsAutoCreatedClaim` — unit tests for the helper: both markers required, missing label → false, missing annotation → false
- [x] `TestResourceClaimOwnershipController_Reconcile` — confirms manual claims (including partial-marker cases) return `ctrl.Result{}, nil` without touching the claim; auto-created claims proceed past the guard to `resolveOwner`
- [x] Existing `TestDeniedAutoClaimCleanupController` — all 6 cases still pass after the cleanup controller refactor

Closes #642

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/internal/quota/controllers/lifecycle/cleanup.go
+++ b/internal/quota/controllers/lifecycle/cleanup.go
@@ -157,15 +157,11 @@ func (r *DeniedAutoClaimCleanupController) stalePendingAge() time.Duration {
 // isAutoCreatedClaim checks if a ResourceClaim was automatically created by the admission plugin.
 // Returns true only if both the label and annotation markers are present.
 func (r *DeniedAutoClaimCleanupController) isAutoCreatedClaim(claim *quotav1alpha1.ResourceClaim) bool {
-	autoCreatedLabel := claim.Labels["quota.miloapis.com/auto-created"] == "true"
-	createdByPlugin := claim.Annotations["quota.miloapis.com/created-by"] == "claim-creation-plugin"
-
 	r.logger.V(3).Info("Checking auto-created markers",
 		"claim", claim.Name,
-		"autoCreatedLabel", autoCreatedLabel,
-		"createdByPlugin", createdByPlugin)
-
-	return autoCreatedLabel && createdByPlugin
+		"autoCreatedLabel", claim.Labels["quota.miloapis.com/auto-created"] == "true",
+		"createdByPlugin", claim.Annotations["quota.miloapis.com/created-by"] == "claim-creation-plugin")
+	return isAutoCreatedClaim(claim)
 }
 
 // isClaimDenied checks if a ResourceClaim has been denied due to quota exceeded.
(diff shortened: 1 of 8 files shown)


drewr (approved):
🛸

Issue #645, opened 2026-06-04
A restart loop in production stalls PolicyBinding reconciliation


ecv:
## Summary

On **prod** (`datum-system/milo-controller-manager`, image `ghcr.io/milo-os/milo:v0.27.0`, single replica) the manager is in a **restart + cold-resync loop** and never reaches steady state. Because the PolicyBinding controller runs in this same manager, newly-created `PolicyBinding`s are **never reconciled** — they get the `iam.miloapis.com/policybinding` finalizer but **no `.status`, no conditions, no events** — so no authz tuples are written and the grant never takes effect.

Discovered while standing up IAM for the inventory service: 6 freshly-created `PolicyBinding`s sat statusless for 17+ min and every `inventory.miloapis.com` verb stayed `Forbidden`, despite correct Roles/ProtectedResources/bindings.

## Evidence

Pod `milo-controller-manager-75dbb6d99b-6ghp8`:
- `restartCount: 11`, last state `terminated, exitCode: 2, reason: Error`
- container mem limit 10Gi / request 6Gi (so not obviously the OOM in #631, but adjacent)

Startup — leader-lease GET to apiserver times out repeatedly:
```
leaderelection: Error retrieving lease lock  lock="milo-system/datum-controller-manager"
  err="Get .../leases/datum-controller-manager?timeout=5s: Client.Timeout exceeded while awaiting headers"
Failed to acquire lease
```

Once leader, it's buried in a full cold-resync across **all** project control planes — quota grant-creation (`default-{networking,notes,dns,core}-quota-policy`) per project + GC backlog + client-side throttling (`Waited before sending request ... reason: client-side throttling`). Recurring GC error:
```
garbage-collector-controller: error syncing item ... err="unable to get REST mapping for iam.miloapis.com/User."
```
It appears to restart (exit 2) before draining this, so it loops.

Affected `PolicyBinding` (one of six, all identical shape):
```
spec.resourceSelector.resourceKind: {apiGroup: inventory.miloapis.com, kind: Site}
spec.roleRef: {name: inventory.miloapis.com-editor, namespace: milo-system}
spec.subjects: [{kind: User, name: evetere@datum.net, uid: "339018957535055630"}]
metadata.finalizers: [iam.miloapis.com/policybinding]
status: <absent>      # 17+ min after creation
```

## Impact

- IAM `PolicyBinding` reconciliation is stalled platform-wide on prod → new grants don't take effect.
[...]

PR #646, opened 2026-06-04
fix(controller-manager): exit deterministically instead of panicking on manager shutdown

ecv:
## Problem

Addresses the exit-2 / leader-lease crashloop in **#645** (and the recurring pathology in **#631**).

The leader callback (`OnStartedLeading`) starts three long-lived runnables in goroutines:

```go
go func() { if err := mcMgr.Start(ctx); err != nil { logger.Error(...); panic(err) } }()
go func() { if err := infraCluster.Start(ctx); err != nil { panic(err) } }()
go func() { if err := ctrl.Start(ctx); err != nil { panic(err) } }()
```

controller-runtime's `Manager.Start` (and the cluster / multicluster-runtime `Start`) returns a **non-nil error when its runnables miss the graceful-shutdown deadline**. Under the prod cold-resync storm (quota grant-creation across all project control planes + GC backlog + client-side throttling), that's exactly what happens on leader-context cancellation. The `panic(err)` then:

- terminated the process with **exit code 2** (matching the reported `exitCode: 2`), with a panic stack that **obscured the real cause** (missed grace deadline vs. OOM vs. context error), and
- exited **mid-shutdown, skipping clean leader-lease release** — so the next pod's lease GET timed out until the lease expired (`leaderelection: Error retrieving lease lock ... Client.Timeout exceeded`), feeding the observed restart + cold-resync loop (`restartCount: 11`).

## Change

Replace the three runtime `panic(err)` calls with logged `klog.FlushAndExit(klog.ExitFlushTimeout, 1)` — the same deterministic error handling used everywhere else in this callback. The process now exits with a **consistent code (1)**, **flushes the real error** to the logs, and terminates the sibling managers so the lease is released promptly.

## Scope / what this does NOT fix

This makes the failure **deterministic and diagnosable** and stops the panic-driven crashloop, but it does **not** eliminate the underlying cold-resync storm that triggers the overloaded shutdown. Those need separate, metrics-informed work (tracked for follow-up):

1. **Quota grant-creation parallelism** — `default-{networking,notes,dns,core}-quota-policy` created per project control plane on every cold start; consider bounding concurrency / backoff so a cold start doesn't saturate the client (QPS 500 / Burst 1000 on the quota dynamic client today).
2. **GC `unable to get REST mapping for iam.miloapis.com/User`** — the garbage collector keeps trying to resolve the cluster-scoped `User` owner kind in project control planes that don't serve it, churning the GC backlog. Likely wants the unmappable cluster-scoped IAM kinds excluded from per-project GC.
3. **Leader-election timing vs. throttling** — under client-side throttling the lease renew can miss `RenewDeadline`, dropping leadership and forcing the (now-clean) shutdown. Worth revisiting lease timing and/or `GracefulShutdownTimeout`, and the single-replica / resource sizing (cf. #631 memory growth).

## Testing

`go build ./cmd/...` passes. The change is process-lifecycle/logging only (no unit-testable logic); behavior verified by reading the controller-runtime shutdown contract (Start returns error on missed grace deadline) against the panic path.

### DIFF
--- a/cmd/milo/controller-manager/controllermanager.go
+++ b/cmd/milo/controller-manager/controllermanager.go
@@ -701,8 +701,13 @@ func Run(ctx context.Context, c *config.CompletedConfig, opts *Options) error {
 			go func() {
 				logger.Info("Starting multicluster manager for quota system")
 				if err := mcMgr.Start(ctx); err != nil {
-					logger.Error(err, "Multicluster manager failed")
-					panic(err)
+					// Exit deterministically (code 1) and flush logs instead of
+					// panicking (which exits 2 with a stack that obscures the
+					// real cause, e.g. a missed graceful-shutdown deadline under
+					// load). FlushAndExit terminates the process, stopping the
+					// sibling managers and releasing the leader lease.
+					logger.Error(err, "Multicluster manager failed; shutting down controller-manager")
+					klog.FlushAndExit(klog.ExitFlushTimeout, 1)
 				}
 			}()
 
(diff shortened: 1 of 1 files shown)


ecv:
It's the little things.

scotwells (approved):

PR #653, opened 2026-06-15
feat(iam): resolve PolicyBinding subject uid from name

### LINKED ISSUE #652: IAM: resolve PolicyBinding Group/User subject uid by name (enable declarative GitOps bindings)
## Summary

Allow a `PolicyBinding` to reference a `Group` subject **by name** and have the IAM controller resolve the Group's `metadata.uid`, instead of requiring callers to supply the uid up front.

## Problem

`PolicyBinding` subject validation (`pkg/apis/iam/v1alpha1/policybinding_types.go`) requires a uid for every non-`system:` subject:

```
(self.kind == 'Group' && has(self.name) && self.name.startsWith('system:')) || (has(self.uid) && size(self.uid) > 0)
```

A Group's `metadata.uid` is cluster-assigned at create time and unknown at commit time. This makes Group bindings **non-declarative**: you cannot commit a `Group` + its `PolicyBinding` to Git and let GitOps apply them, because the binding needs a uid that only exists after the Group reconciles.

Today every caller works around this imperatively — read the uid back, then create the binding:
- Internal controllers do it (`organization_membership`, `userinvitation`, `project_webhook` all set `UID: string(...GetUID())` before creating bindings).
- External GitOps consumers must ship an out-of-band render-after-reconcile script. Concrete case: datum-cloud/infra binds staff/loader Groups to the inventory graph roles and needs `bin/apply-inventory-iam-bindings.sh` to read each Group's uid and `envsubst` it into the binding — purely because of this constraint.

## Proposed

When a `PolicyBinding` subject is `kind: Group` (or `User`/`ServiceAccount`) given by `name`/`namespace` **without** `uid`, have the IAM layer resolve and populate the uid from the named object — the same lookup the internal controllers already do, centralized.

Options (any one solves it):
1. **Mutating/defaulting webhook** on `PolicyBinding` create/update: resolve `subject.name` → `subject.uid` before the validating CEL runs. Fully transparent; bindings become declaratively committable.
2. **Controller-side resolution**: relax the CEL to allow a name-only Group subject, and have the PolicyBinding controller resolve + persist the uid (or carry it only in status / the emitted authz tuples).

Webhook (1) is least disruptive — keeps uid immutable in the stored spec, just fills it from the name at admission.

## Why uid-on-spec exists (and why resolving is safe)

The uid pins the binding to a specific object instance so a delete+recreate of a same-named Group doesn't silently re-grant. Resolving at admission preserves that: the webhook stamps the *current* uid, so a later recreate yields a different uid and the binding no longer matches until re-applied — same guarantee, without pushing the read-back onto every caller.

## Impact

- GitOps-native Group/User bindings: commit `Group` + `PolicyBinding` together, Flux applies them, no render scripts.
- Removes a class of bespoke per-consumer tooling (e.g. the inventory IAM apply script).

## References

- Validation: `pkg/apis/iam/v1alpha1/policybinding_types.go` (subject XValidation rule)
- Existing imperative read-back: `internal/controllers/iam/userinvitation_controller.go`, `internal/webhooks/resourcemanager/v1alpha1/project_webhook.go`
- Downstream workaround: datum-cloud/infra `bin/apply-inventory-iam-bindings.sh` (+ `apps/inventory/staging/milo/*.template.yaml`)

ecv:
## Summary

Closes #652.

Adds a mutating/defaulting webhook on `PolicyBinding` that resolves a subject's `metadata.uid` from its `name` when the uid is omitted. For each subject with no uid, the webhook looks up the named `User`, `Group`, or `ServiceAccount` and stamps its current uid into the subject — before the validating CEL rule runs, so the existing `Subject` validation passes unchanged.

This implements **Option 1** from the issue (least disruptive): the uid stays immutable in the stored spec; admission just fills it from the name.

## Why

A `Group`'s `metadata.uid` is cluster-assigned at create time and unknown at commit time, which made Group/User bindings **non-declarative** — every caller had to read the uid back imperatively (internal controllers do; external GitOps consumers shipped render-after-reconcile scripts). Now a `Group` + its `PolicyBinding` can be committed and applied together.

## Behavior

- **Resolves** `User` (cluster-scoped, by name), `ServiceAccount` (cluster-scoped, by name), and `Group` (namespaced, by name + namespace).
- **Skips** subjects that already carry a uid — uid remains immutable; internal controllers that set it explicitly are unaffected.
- **Skips** system groups (`name` starting with `system:`) — no backing object, matching the existing CEL exemption.
- **Rejects** with a clear admission error when the named subject does not exist, or when a `Group` subject omits its namespace.

### Instance-pinning preserved

The webhook stamps the *current* uid, so a delete+recreate of a same-named subject yields a different uid and the binding no longer matches until re-applied — same guarantee the uid-on-spec requirement provided, without pushing the read-back onto every caller.

## Changes

- `internal/webhooks/iam/v1alpha1/policybinding_webhook.go` — `PolicyBindingMutator` + `SetupPolicyBindingWebhooksWithManager`.
- `cmd/milo/controller-manager/controllermanager.go` — wire the webhook into the controller manager.
- `config/webhook/manifests.yaml` — generated mutating webhook config (`mpolicybinding.iam.miloapis.com`, CREATE/UPDATE).
- `config/controller-manager/overlays/core-control-plane/rbac/role.yaml` — generated; grants `get;list;watch` on `serviceaccounts` (users/groups already present).
- `Taskfile.yaml` — include `./internal/webhooks/...` in the RBAC generation paths so webhook `+kubebuilder:rbac` markers are honored.

## Testing

- **Unit:** `internal/webhooks/iam/v1alpha1/policybinding_webhook_test.go` — 8 subtests (User/Group/ServiceAccount resolution, multi-subject, skip-when-set, skip system group, not-found error, missing-namespace error). `go test ./internal/webhooks/iam/...` passes.
- **E2E (Chainsaw):** `test/iam/policybinding-subject-uid-resolution/` — name-only Group subject gets its uid resolved; non-existent Group is rejected at admission; system group admitted without a uid. Test document validates with `chainsaw lint`.
- `go build ./...` clean.

### DIFF
--- a/internal/webhooks/iam/v1alpha1/policybinding_webhook.go
+++ b/internal/webhooks/iam/v1alpha1/policybinding_webhook.go
@@ -122,7 +122,9 @@ func (m *PolicyBindingMutator) resolveSubjectUID(ctx context.Context, subject *i
 // an admission response.
 func lookupFieldError(namePath *field.Path, name, kind string, err error) *field.Error {
 	if errors.IsNotFound(err) {
-		return field.Invalid(namePath, name, fmt.Sprintf("%s %q not found; cannot resolve subject uid", kind, name))
+		notFound := field.NotFound(namePath, name)
+		notFound.Detail = fmt.Sprintf("%s %q does not exist; cannot resolve subject uid", kind, name)
+		return notFound
 	}
 	return field.InternalError(namePath, fmt.Errorf("failed to get %s %q: %w", kind, name, err))
 }
--- a/internal/webhooks/iam/v1alpha1/policybinding_webhook.go
+++ b/internal/webhooks/iam/v1alpha1/policybinding_webhook.go
@@ -1,0 +1,128 @@
+package v1alpha1
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+
+	iamv1alpha1 "go.miloapis.com/milo/pkg/apis/iam/v1alpha1"
+)
+
+// systemGroupPrefix identifies Group subjects that are synthesized by the
+// platform (for example "system:authenticated-users"). These groups have no
+// backing object and therefore no uid to resolve.
+const systemGroupPrefix = "system:"
+
+func SetupPolicyBindingWebhooksWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewWebhookManagedBy(mgr, &iamv1alpha1.PolicyBinding{}).
+		WithDefaulter(&PolicyBindingMutator{
+			client: mgr.GetClient(),
+		}).
+		Complete()
+}
+
+// +kubebuilder:webhook:path=/mutate-iam-miloapis-com-v1alpha1-policybinding,mutating=true,failurePolicy=fail,sideEffects=None,groups=iam.miloapis.com,resources=policybindings,verbs=create;update,versions=v1alpha1,name=mpolicybinding.iam.miloapis.com,admissionReviewVersions={v1,v1beta1},serviceName=milo-controller-manager,servicePort=9443,serviceNamespace=milo-system
+
+// +kubebuilder:rbac:groups=iam.miloapis.com,resources=users;groups;serviceaccounts,verbs=get;list;watch
+
+// PolicyBindingMutator resolves the uid of each PolicyBinding subject from its
+// name. Callers may reference a User, Group, or ServiceAccount by name without
+// supplying a uid; the mutator looks up the named object and stamps its current
+// uid into the subject. This makes bindings declaratively committable (a Group
+// and its PolicyBinding can be applied together by GitOps) while preserving the
+// instance-pinning guarantee: the stored uid still identifies one specific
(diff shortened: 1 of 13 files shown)


ecv:
Adding chainsaw now…

ecv:
This lets us greatly simplify https://github.com/datum-cloud/infra/pull/2706

ecv:
bump

internal/webhooks/iam/v1alpha1/policybinding_webhook.go
```diff
@@ -0,0 +1,128 @@
...
+	default:
+		return "", field.NotSupported(subjectPath.Child("kind"), subject.Kind, []string{"User", "Group", "ServiceAccount"})
+	}
+}
+
+// lookupFieldError converts a client.Get error into a field error suitable for
+// an admission response.
+func lookupFieldError(namePath *field.Path, name, kind string, err error) *field.Error {
+	if errors.IsNotFound(err) {
+		return field.Invalid(namePath, name, fmt.Sprintf("%s %q not found; cannot resolve subject uid", kind, name))
```
JoseSzycho: a `field.NotFound(....` returning error is going to be more accurate

JoseSzycho (changes requested):
Looks great, just left a minor review

internal/webhooks/iam/v1alpha1/policybinding_webhook.go
ecv: Good call — switched to `field.NotFound` (kind context preserved in the error detail) in 373ae33.

JoseSzycho (approved):

Issue #659, opened 2026-06-17
iam: migrate PlatformAccessApproval and PlatformAccessRejection to PlatformAccess in milo and fraud


JoseSzycho:
## Context

Part of milo-os/milo#613 — replacing `UserDeactivation`, `PlatformAccessApproval`, and `PlatformAccessRejection` with a single mutable `PlatformAccess` resource.

This issue covers migrating the `PlatformAccessApproval` and `PlatformAccessRejection` resources and all systems that produce or consume them.

## Background

`PlatformAccessApproval` and `PlatformAccessRejection` are immutable, write-once resources that govern `User.status.registrationApproval` (Pending/Approved/Rejected). Moving a user from approved to rejected today requires deleting the approval and creating a rejection — two operations with a window of inconsistency. Under the new model, the `Approved` and `Rejected` states on `PlatformAccess` replace both resources with a single mutable update.

### milo-os/milo

- [ ] Update `UserController` to derive registration state from `PlatformAccess.spec.state` instead of watching `PlatformAccessApproval` and `PlatformAccessRejection`
- [ ] Remove the `PlatformAccessApproval` controller
- [ ] Remove the `PlatformAccessRejection` controller
- [ ] Update `User.status` to reflect `accessState` derived from `PlatformAccess` (remove `registrationApproval`; add `accessState`) — coordinate with milo-os/milo#658
- [ ] Remove the `PlatformAccessApproval` and `PlatformAccessRejection` CRDs and all associated RBAC, roles, and protected resource config once no consumers remain
- [ ] Update e2e tests that exercise the approval and rejection flows

### milo-os/fraud

- [ ] Update the fraud operator to create or update a `PlatformAccess` resource (setting `state: Approved` or `state: Rejected`) instead of creating `PlatformAccessApproval` or `PlatformAccessRejection` resources
- [ ] Update the operator to set `state: Suspended` on `PlatformAccess` instead of creating `UserDeactivation` resources (coordinate with milo-os/milo#658)
- [ ] Reference milo-os/fraud#42 for full context on the fraud operator changes

## Migration Notes

- `PlatformAccessApproval` creation maps to setting `PlatformAccess.spec.state = Approved`
- `PlatformAccessRejection` creation maps to setting `PlatformAccess.spec.state = Rejected`
- Transitioning from approved to rejected (previously a delete + create) is now a single update to `PlatformAccess.spec.state`

## Related

- milo-os/milo#613 — parent issue and full design
[...]

Issue #658, opened 2026-06-17
iam: migrate UserDeactivation to PlatformAccess in milo and zitadel-provider


JoseSzycho:
## Context

Part of #613 — replacing `UserDeactivation`, `PlatformAccessApproval`, and `PlatformAccessRejection` with a single mutable `PlatformAccess` resource.

This issue covers migrating the `UserDeactivation` resource and all systems that produce or consume it.

## Background

`UserDeactivation` currently governs `User.status.state` (Active/Inactive). The `UserController` in Milo watches these resources to set that status field, and the Zitadel provider watches them to activate or deactivate users in Zitadel. Under the new model, the `Suspended` state on `PlatformAccess` replaces this resource entirely.

### milo-os/milo

- [x] Update `UserController` to watch `PlatformAccess` for `state: Suspended` instead of `UserDeactivation`
- [x] Remove the `UserDeactivation` controller
- [x] Update `User.status` to reflect `accessState` derived from `PlatformAccess` (remove `state`, `registrationApproval`; add `accessState`)
- [x] Update the `Ready` condition semantics on `User` so `Ready=True` only when `accessState=Approved`
- [x] Remove the `UserDeactivation` CRD and all associated RBAC, roles, and protected resource config once no consumers remain
- [x] Update e2e tests that exercise the deactivation flow

### milo-os/zitadel-provider

- [x] Update the provider to watch `PlatformAccess` for `state: Suspended` and `state: Approved` (reactivation) instead of watching `UserDeactivation` creates and deletes
- [x] Ensure the transition `Suspended → Approved` reactivates the user in Zitadel, equivalent to deleting a `UserDeactivation` today

## Migration Notes

- `UserDeactivation` delete (reactivation) maps to setting `PlatformAccess.spec.state = Approved`
- `UserDeactivation` create (deactivation) maps to setting `PlatformAccess.spec.state = Suspended`
- Any system creating `UserDeactivation` resources (e.g. the fraud operator) is covered in a separate issue

## Related

- #613 — parent issue and full design
- milo-os/fraud#42 — fraud operator migration (covered separately)

JoseSzycho:
All PR has been merged and everything works correctly in staging.

PR #660, opened 2026-06-17
feat(iam): Add PlatformAccess resource for unified user access lifecycle

### LINKED ISSUE #613: Simplify user access lifecycle with a single PlatformAccess resource
## Background

Today, governing whether a user can access the platform requires three separate resources: `UserDeactivation`, `PlatformAccessApproval`, and `PlatformAccessRejection`. Each was introduced incrementally to handle a specific scenario, but together they create a model that is difficult to reason about and awkward to operate.

## Problems

**Reactivating a suspended user is unintuitive.** When the fraud operator deactivates a user, it creates a `UserDeactivation` resource. To reactivate the user, an admin must know to delete that resource — there is no "reactivate" action, just a delete of an object the fraud system created.

**Disapproving an approved user requires two steps.** `PlatformAccessApproval` and `PlatformAccessRejection` are mutually exclusive and immutable. To move a user from approved to rejected, an admin must first delete the approval and then create a rejection — two operations with a window of inconsistency between them.

**The model has no lifecycle.** All three resources are effectively write-once. There is no way to transition a user between states without deleting and recreating resources, which makes operational tooling and audit reasoning harder.

**User state is split across two axes.** `User.status.state` (Active/Inactive) and `User.status.registrationApproval` (Pending/Approved/Rejected) are governed by different resources but answer the same question: can this user access the platform?

**User status is misleading.** Today a user in `registrationApproval: Pending` — who cannot access the platform — will also show `Ready=True` on their status conditions, because `Ready` reflects whether the controller successfully reconciled the resource, not whether the user has functional access. This creates a confusing signal for API consumers, dashboards, and CLI tooling.

## Proposal

Replace `UserDeactivation`, `PlatformAccessApproval`, and `PlatformAccessRejection` with a single mutable `PlatformAccess` resource — one per user — that owns the full access lifecycle.

```yaml
apiVersion: iam.miloapis.com/v1alpha1
kind: PlatformAccess
metadata:
  name: <user-name>
spec:
  userRef:
    name: <user-name>
  state: Pending | Approved | Rejected | Suspended
  reason: <string>
```

The `UserController` derives all access state from this single resource. `User.status.registrationApproval` and `User.status.state` collapse into a single `User.status.accessState` mirroring `PlatformAccess.spec.state`.

The `Ready` condition on `User` is updated to reflect whether the user has functional platform access, not just whether the controller reconciled successfully:

- `Ready=True` — only when `accessState=Approved`
- `Ready=False` — for all other states, with a meaningful reason string:
  - `reason: RegistrationPending`
  - `reason: Rejected`
  - `reason: Suspended`

This means any API consumer, dashboard, or CLI only needs to check one field to answer "can this user access the platform," and gets a semantically correct answer.

**State machine:**
- `Pending → Approved` — fraud accepts, or admin approves
- `Pending → Rejected` — fraud or admin rejects
- `Approved → Suspended` — fraud deactivates, or admin suspends
- `Suspended → Approved` — admin reactivates
- `Approved → Rejected` — admin disapproves

Audit history (who changed the state and why) is captured in the API server audit log rather than embedded in resource spec fields, keeping the resource focused on current state.

## Motivation

- Single resource to inspect when asking "why can't this user access the platform"
- State transitions are updates, not delete-then-create operations
- External systems (e.g. the fraud operator) create and update a single resource rather than managing three
- Simpler `UserController` — one watch, one read, one status field to set
- `Ready` condition accurately reflects whether the user has platform access, eliminating misleading status signals

## Open Questions

- Should `PlatformAccess` be created eagerly on user creation (every user always has one) or lazily on first action?
- Is `Suspended` meaningfully distinct from `Rejected` in terms of user communication and reversibility, or should they be unified?

## Related

- milo-os/fraud#42 — fraud operator changes required to use `PlatformAccess` for enforcement

JoseSzycho:
## Summary

Today, governing whether a user can access the platform requires three separate resources — `UserDeactivation`, `PlatformAccessApproval`, and `PlatformAccessRejection` — each immutable and each answering only part of the question. Moving a user between states requires delete-then-create sequences with windows of inconsistency, and there is no single place to look to understand why a user cannot access the platform.

This PR introduces `PlatformAccess`, a single mutable cluster-scoped resource that owns the full access lifecycle for a user. It is the first step in the migration described in #613 and intentionally introduces only the API type; controller logic, webhook updates, and removal of the existing resources are tracked in follow-on issues.

> [!NOTE]
> The existing `UserDeactivation`, `PlatformAccessApproval`, and `PlatformAccessRejection` resources are untouched in this PR. Migration of controllers, the authn webhook, and the `ValidatingAdmissionPolicy` in `datum-cloud/datum` is tracked in #658, #659, and a separate issue for the admission policy update.


mattdjenkinson (approved):
We can hook the work done for https://github.com/datum-cloud/enhancements/pull/779 & https://github.com/datum-cloud/enhancements/pull/770 into this.

PR #670, opened 2026-06-26
refactor(quota): make quota admission plugin reusable by aggregated API servers

scotwells:
## What & why

Makes Milo's quota admission plugin **reusable by other aggregated API servers**, so services beyond `milo-apiserver` can enforce Milo quota in their own request path.

The first consumer is the IPAM service, which needs to protect finite IP address space by gating consumption per project. Because IPAM is an *aggregated* API server, quota can't be enforced from `milo-apiserver`'s admission chain (aggregated requests are proxied straight to the backend) — IPAM must run the quota plugin itself. This PR extracts that plugin into an importable library so IPAM (and future services) can register it without copying code.

Tracked by **milo-os/ipam#39**.

## Change

Pure refactor — **no behavior change for Milo**. Moves the quota admission plugin and its dependencies out of `internal/` into an importable `pkg/quota/...` library, and removes the two couplings that prevented external import:

- **Move** `internal/quota/{admission,engine,validation,cel,templateutil}` → `pkg/quota/...` (git renames, history preserved). `internal/quota/controllers/` stays internal.
- **Drop the `k8s.io/kubernetes` dependency** from the plugin: the single `legacyscheme.Scheme.ConvertToVersion` call (only reached for structured native types, never for CRDs) is replaced by an injected, nil-guarded `runtime.ObjectConvertor`. Milo injects `legacyscheme.Scheme` from its `cmd/` wiring, so native-type behavior is unchanged; CRD-only hosts leave it nil.
- **Drop the `pkg/server/filters` dependency** from the plugin: `OrganizationID` / `WithOrganization` move into `pkg/request` (single source of truth for request context); `pkg/server/filters` now reads through it.
- Repoint all of Milo's own call sites (`cmd/milo/apiserver/*`, `internal/quota/controllers/*`).

## Public API exposed for reuse

- `go.miloapis.com/milo/pkg/quota/admission` — `PluginName`, `Register(*admission.Plugins)`, `ReadinessCheck()`, and dependency setters (`SetLoopbackConfig`, `SetObjectConvertor`, dynamic client via the stock `WantsDynamicClient` initializer).
- `go.miloapis.com/milo/pkg/request` — `WithProject`/`ProjectID`/`WithOrganization`/`OrganizationID`.

The moved packages have **zero `internal/` imports and zero `k8s.io/kubernetes` import**, so they're cleanly importable from another module.

## Verification

- `go build ./...` — clean
- `go vet ./...` — clean
- `go test ./pkg/quota/... ./pkg/request/... ./pkg/server/filters/... ./internal/quota/... ./cmd/...` — pass
- `grep -rn "go.miloapis.com/milo/internal/" pkg/quota/` → none
- `grep -rn "k8s.io/kubernetes" pkg/quota/` → none (imports)

## Notes for reviewers

- Draft until reviewed.
- Milo continues to register the plugin exactly as before — the only Milo-side addition is injecting `legacyscheme.Scheme` via the new `ObjectConvertor` initializer.
- IPAM-side wiring (registering the plugin, quota client, service-catalog registration, grant-on-enablement) is **not** in this PR; it's tracked in milo-os/ipam#39.


ecv (approved):
hell yeah scot's back on ipam baby

PR #672, opened 2026-06-27
fix: null-guard responseObject.metadata.name in ActivityPolicy creates

ecv:
## What

Make the `create` rule summaries null-safe in six ActivityPolicies that
dereference `audit.responseObject.metadata.name`. Guard the leaf and fall back
to `audit.objectRef.name`.

## Why

Same DLQ-leak class as the prod `DLQSlowLeak` on the NSO gateway policy
(milo-os/activity#212). A `create` rule matches on verb/requestObject, so a
**rejected** create (409/422/admission-deny) still matches — but the apiserver
returns a `Status` object as `responseObject` with no `metadata.name`. CEL
raises `no such key: name`, the event goes to the DLQ, and retries fail
identically. Latent until the first rejected create of each kind.

## Fixed (create rules)

- `config/services/activity/policies/resourcemanager/` — project, organization
- `config/services/activity/policies/iam/` — role, group, serviceaccount
- `config/services/identity/policies/` — serviceaccount

## Remediation

Merge + milo release → the `milo-activity-policies` Flux Kustomization
re-applies the corrected CRs; processor retries any DLQ backlog. No `kubectl`.

## Related

- NSO fix (gateway + 5 more): datum-cloud/network-services-operator#230
- billing fix: milo-os/billing (separate PR)
- Runbook fix: milo-os/activity#213 · Investigation: milo-os/activity#212

### DIFF
--- a/config/services/activity/policies/iam/group-policy.yaml
+++ b/config/services/activity/policies/iam/group-policy.yaml
@@ -25,7 +25,7 @@ spec:
   auditRules:
     - name: create
       match: "!audit.user.username.startsWith('system:') && audit.verb == 'create'"
-      summary: "{{ actor }} created group {{ has(audit.responseObject.metadata.name) ? link(audit.responseObject.metadata.name, audit.objectRef) : link(audit.objectRef.name, audit.objectRef) }}"
+      summary: "{{ actor }} created group {{ has(audit.responseObject.metadata.name) ? link(audit.responseObject.metadata.name, audit.objectRef) : has(audit.objectRef.name) ? link(audit.objectRef.name, audit.objectRef) : (has(audit.responseObject.details) && has(audit.responseObject.details.name)) ? link(audit.responseObject.details.name, audit.objectRef) : link('a group', audit.objectRef) }}"
 
     - name: delete
       match: "!audit.user.username.startsWith('system:') && audit.verb == 'delete'"
--- a/config/services/activity/policies/iam/group-policy.yaml
+++ b/config/services/activity/policies/iam/group-policy.yaml
@@ -25,7 +25,7 @@ spec:
   auditRules:
     - name: create
       match: "!audit.user.username.startsWith('system:') && audit.verb == 'create'"
-      summary: "{{ actor }} created group {{ link(audit.responseObject.metadata.name, audit.objectRef) }}"
+      summary: "{{ actor }} created group {{ has(audit.responseObject.metadata.name) ? link(audit.responseObject.metadata.name, audit.objectRef) : link(audit.objectRef.name, audit.objectRef) }}"
 
     - name: delete
       match: "!audit.user.username.startsWith('system:') && audit.verb == 'delete'"
--- a/config/services/activity/policies/iam/role-policy.yaml
+++ b/config/services/activity/policies/iam/role-policy.yaml
@@ -12,7 +12,7 @@ spec:
   auditRules:
     - name: create
       match: "!audit.user.username.startsWith('system:') && audit.verb == 'create'"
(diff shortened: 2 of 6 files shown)


ecv:
### Blast radius + suggested follow-up (not in this PR)

While extending this fix (added `responseObject.details.name` fallback for `generateName` creates rejected before naming — see latest commit), I swept the rest of the activity policies for the same `responseObject` deref bug class. Summary of what's left:

**Already safe** — these non-create rules deref `responseObject` but their `match` clause guards it down to the ref object, so a rejected/Status response never reaches the summary:

- `resourcemanager/organizationmembership` (delete) — `has(responseObject.spec) && has(responseObject.spec.userRef)`
- `notification/contactgroupmembership` (delete) — guards `contactRef` + `contactGroupRef`
- `iam/groupmembership` (delete) — guards `userRef` + `groupRef`
- `iam/userdeactivation` (delete) — guards `userRef`
- `quota/resourcegrant` (delete) — guards `consumerRef`
- `iam/userinvitation` (accept/decline/cancel) — guards `responseObject.status.organization.displayName` fully → safe

**Residual (low-probability) gap:** each of the above derefs **one level deeper** than its `match` guards — e.g. `match` checks `...spec.userRef`, summary uses `...spec.userRef.name`. If the ref object is ever present without `name`, CEL raises `no such key: name` → DLQ. In practice refs always carry a name, so this is latent, not active. No DLQ evidence for these.

**Recommendation:** a separate follow-up to push each guard to the leaf (`&& has(...userRef.name)`) or fall the summary back to `objectRef.name`. Out of scope for this PR — flagging so it's tracked, **not pursuing it here**.

JoseSzycho (approved):

scotwells:
@ecv can you help me understand what the audit logs for these were? The metadata name should always be present in the audit logs.

ecv:
@scotwells here are the actual audit logs from the prod activity processor (`activity-system/activity-processor`, core control plane). They explain why `responseObject.metadata.name` is missing.

The `metadata.name` **is** always present on a successful response object. But a **rejected** create returns a `Status` object as `responseObject`, and `Status.metadata` is the empty `ListMeta` — `metadata: {}` — so there is no `name` key. The `create` rule still matches (it matches on verb/requestObject), the summary derefs `responseObject.metadata.name`, and CEL raises `no such key: name`.

Concrete prod example (project quota 403 — auditID `a6d769ee-6417-4620-893a-9364d8b809f9`):

```json
{
  "auditID": "a6d769ee-6417-4620-893a-9364d8b809f9",
  "verb": "create",
  "level": "RequestResponse",
  "requestURI": "/apis/resourcemanager.miloapis.com/v1alpha1/projects",
  "objectRef": {
    "apiGroup": "resourcemanager.miloapis.com",
    "name": "personal-project-30ea0ba6",
    "resource": "projects"
  },
  "responseObject": {
    "apiVersion": "v1",
    "kind": "Status",
    "status": "Failure",
    "reason": "Forbidden",
    "code": 403,
    "message": "projects.resourcemanager.miloapis.com \"personal-project-30ea0ba6\" is forbidden: You've reached your quota for this resource type ...",
    "metadata": {},
    "details": { "group": "resourcemanager.miloapis.com", "kind": "projects", "name": "personal-project-30ea0ba6" }
  },
  "responseStatus": { "code": 403, "reason": "Forbidden", "status": "Failure", "metadata": {} }
}
```

Processor error logged alongside it:

```
rule 0 summary: summary template evaluation failed:
eval "link(audit.responseObject.metadata.name, audit.objectRef)": no such key: name
```

This exact shape repeats across many auditIDs from real users hitting project quota and retrying the create. Note that the name is still recoverable from `objectRef.name` (`personal-project-30ea0ba6`) and `responseObject.details.name` even when `responseObject.metadata.name` is absent — which is what the `objectRef.name` fallback in this PR relies on. So the right general guard, as you suggested, is to handle failed requests (`responseStatus.code >= 300` / `responseObject.kind == "Status"`) rather than assume `metadata.name`.

Issue #636, opened 2026-06-01
RFC: Collapse Personal/Standard org types into one type, tag business separately


mattdjenkinson:
## Summary

`Organization.Type` is today an **immutable** enum — `Personal | Standard` — that doubles as both a tenancy distinction and a proxy for "is this a serious/business account." This RFC asks whether we should **collapse those two types into a single org type** and instead represent "business" as a **mutable tag/attribute**, not a structural type.

**Core insight: business-vs-individual is a billing/tax property, not a tenancy property.** An org is an org — a workspace that holds projects and members. Whether the entity behind it is a business (with a legal name and tax ID) is an attribute that can change over time and belongs alongside billing identity, not baked into an immutable tenancy type.

## Motivation

The current `Personal`/`Standard` split causes real friction:

- **Immutability trap.** `Type` is immutable, and `Personal` orgs additionally have an immutable display name. A solo developer who later incorporates cannot convert their org — they must create a new `Standard` org and abandon their workspace.
- **Conflates tenancy with billing identity.** "Personal" is being used to mean both "auto-created default workspace" and "not a business." Those are different things.
- **Auto-created personal orgs add clutter and special-case UX.** They sort first, carry a badge, can't be renamed, and are created before the user has expressed any intent.
- **Two code paths for one resource.** Controllers, validation policies, the membership status cache, and the portal all branch on `Type` for what is essentially the same object.

### Goals

- A single, uniform organization type — every org is just an org (mutable name, normal lifecycle).
- "Business" is representable and **mutable**, so an org can become (or stop being) a business account without recreating anything.
- Onboarding and billing can still differentiate a business to collect legal name + tax ID.
- Existing `Personal` and `Standard` orgs migrate cleanly with no loss of workspace.

### Non-Goals

- Final CEL/schema design (defer to implementation).
- Stripe / tax-engine integration specifics.
- Changing the *default tenant on signup* behavior beyond what collapsing the type requires (see Open Questions).

## Current State (verified)

- **Org model** (`pkg/apis/resourcemanager/v1alpha1/organization_types.go`): `OrganizationSpec.Type` ∈ `{Personal, Standard}`, **immutable** (CEL). `Personal` orgs also have an **immutable display name**.
- **Auto-creation** (`datum` repo, `internal/controller/resourcemanager/personal_organization_controller.go`): a controller watches `User` and auto-creates a `Personal` org (`personal-org-{hash}`), an owner `OrganizationMembership`, and a default `Project`.
[...]

PR #675, opened 2026-06-29
feat(resourcemanager): Unify organization model and onboarding gate

mattdjenkinson:
## Summary

Personal and Standard org types split the same resource into two code paths and locked people into workspaces they could not grow out of. This change removes `spec.type`, adds org-level `contactInfo`, and introduces a single onboarding gate the API can enforce.

Portal clients create orgs with `generateName: org-` and a display name only. The mutating webhook defaults that prefix for non-admin callers; the validating webhook rejects user-chosen slugs. An org is onboarding-complete when contact info (email and name), a billing account in the org namespace, and `DefaultPaymentMethodReady` on that account are all present. The organization controller writes that outcome to `status.conditions[OnboardingComplete]`.

Related enhancement: [datum-cloud/enhancements#770](https://github.com/datum-cloud/enhancements/pull/770). Portal UI work is tracked separately and is not in this PR.

Supersedes #673 (reopened after branch restore).

## Test plan

- [x] `go build ./...`
- [x] `go test ./internal/controllers/resourcemanager/... ./internal/webhooks/resourcemanager/...`
- [x] Chainsaw suites updated for org fixtures without `spec.type`
- [ ] Manual create with `generateName: org-` assigns an opaque name and owner membership
- [ ] `OnboardingComplete` flips to True when contact info and billing payment method are ready

## Breaking changes

- `Organization.spec.type` is removed. Clients must stop sending or reading it.
- `OrganizationMembership.status.organization.type` is removed; `contactEmail` is cached instead.
- Non-admin org creates must use `metadata.generateName` with prefix `org-`.

## Notes for reviewers

- Depends on `go.miloapis.com/billing` for typed reads in the onboarding reconciler.
- Migration steps live in `docs/runbooks/unified-organizations-migration.md`.
- Deploy before or with the matching datum PR that removes auto-provision.

Related to #636


scotwells:
The tight coupling this introduces between the billing service and milo's core system seems rigid. Wonder if we can consider a more dynamic model where services can register onboarding gates with an organization that they're responsible for providing status for?

scotwells (dismissed):

JoseSzycho (dismissed):
I'd suggest to review the logic for a last organization member in a separate PR.  We need to prevent the last member to be remove from the ORG, and possible review the scenario in which the Org creator leaves the organization, as it should change the owner of it.

scotwells (approved):

PR #677, opened 2026-06-30
feat(iam): re-home audit-log-querier inheritance to activity service

### LINKED ISSUE #676: Move audit-log-querier role inheritance from core CRD bundle to activity service
## Summary

Re-home the `iam-user-self-manage` → `activity.miloapis.com-audit-log-querier` role inheritance into the **activity service** config, so the foundational milo control-plane layer no longer depends on the activity stack being up.

## Context

In the infra repo, `milo-core-control-plane-crds` (the Flux Kustomization that applies `./crd/overlays/core-control-plane` from the milo kustomize bundle) carries an inline patch:

```yaml
# Allow users to query their own audit logs from the system.
- patch: |
    kind: Role
    metadata: { name: iam-user-self-manage }
    spec:
      inheritedRoles:
        - name: activity.miloapis.com-audit-log-querier
```

Because that inline patch references `activity.miloapis.com-audit-log-querier`, the core CRD/role bundle effectively had to wait for the activity APIService/roles to exist — an **inverted dependency** (activity rolls out *after* the core control plane). It forced a cross-domain `dependsOn` from the milo CP foundation onto the whole activity stack, which pulled the foundation into the infra incident datum-cloud/infra#2939.

`inheritedRoles` is `x-kubernetes-list-type: map` keyed by `name` (per the Role CRD), not an atomic list. So the entry can be owned independently under server-side apply: the activity overlay owns the single `audit-log-querier` entry while the core role bundle continues to own `launchStage` and `includedPermissions`. The move is driven by dependency ordering — the inherited role must exist before something references it — not by a field-manager conflict.

The infra-side fix is datum-cloud/infra#2943 (audit) → PR datum-cloud/infra#2953 (Tier 2), which removes the inline patch and the `crds → activity` dependency. That PR is **blocked on this change**: dropping the inline patch alone removes the "users can query their own audit logs" capability, so the inheritance must first ship with the activity service.

## Ask

In the milo kustomize bundle, add the `iam-user-self-manage` → `activity.miloapis.com-audit-log-querier` inheritance to the **activity service** overlay (`./services/activity`, applied by the activity-gated Kustomization), so it is applied after activity is up and owned by a single field manager.

Note: milo's core role overlay (`config/roles/iam-user-self-manage.yaml`) never carried this inheritance — it existed only as the infra-repo inline patch shown above. There is nothing to remove on the milo side; the core layer is already free of activity coupling. The inline-patch removal happens in datum-cloud/infra#2953.

Net effect: the audit-log-querier capability is preserved, shipped with the activity service instead of the core foundation.

## Coordination

- Blocks: datum-cloud/infra#2953 (draft until this lands).
- Part of: datum-cloud/infra#2943, datum-cloud/infra#2939.
- After this merges and a new bundle is published: un-draft #2953, then cold-start validate bootstrap ordering.

ecv:
## Summary

Closes #676.

Moves the `iam-user-self-manage` → `activity.miloapis.com-audit-log-querier` role inheritance out of the foundational control-plane layer and into the **activity service** overlay (`config/services/activity`), which is only applied once the activity stack is present. The core role bundle now has **zero activity coupling**.

### Why

The `activity.miloapis.com-audit-log-querier` Role is shipped by the activity service. Referencing it from the core role bundle forced the foundational milo control plane to depend on activity rolling out first — an inverted dependency that pulled the foundation into datum-cloud/infra#2939.

## Changes

- **New:** `config/services/activity/roles/iam-user-self-manage-audit-log-querier.yaml` — a partial `Role` manifest that, under server-side apply, merges a single `inheritedRoles` entry into the `iam-user-self-manage` Role.
- **Edit:** `config/services/activity/kustomization.yaml` — wires the new resource into the activity overlay.

## Validation

- `kustomize build config/services/activity` → rc=0; renders the merged `Role` with the `activity.miloapis.com-audit-log-querier` inherited entry.
- `kustomize build config/roles` → 0 references to `audit-log-querier` (core layer is clean).

## Coordination

- Unblocks: datum-cloud/infra#2953 (un-draft after a new bundle is published, then cold-start validate bootstrap ordering).
- Part of: datum-cloud/infra#2943, datum-cloud/infra#2939.

### DIFF
--- a/config/services/activity/roles/iam-user-self-manage-audit-log-querier.yaml
+++ b/config/services/activity/roles/iam-user-self-manage-audit-log-querier.yaml
@@ -1,20 +1,5 @@
 # SPDX-License-Identifier: AGPL-3.0-only
 
-# Grants users the ability to query their own audit logs by inheriting the
-# activity.miloapis.com-audit-log-querier Role into iam-user-self-manage.
-#
-# This inheritance ships with the activity service overlay rather than the core
-# control-plane role bundle on purpose: the audit-log-querier Role is provided
-# by the activity stack, so referencing it from the foundational layer would
-# force the core control plane to depend on activity being up first (an
-# inverted dependency). Keeping it here means the core role bundle has zero
-# activity coupling, and this entry is only applied once activity is present.
-#
-# Under server-side apply this partial Role merges its single inheritedRoles
-# entry into the iam-user-self-manage Role owned by the core role bundle.
-# inheritedRoles is a map-type list keyed by name, so this overlay owns just
-# this entry while the core bundle continues to own launchStage and
-# includedPermissions.
 apiVersion: iam.miloapis.com/v1alpha1
 kind: Role
 metadata:
--- a/config/services/activity/roles/iam-user-self-manage-audit-log-querier.yaml
+++ b/config/services/activity/roles/iam-user-self-manage-audit-log-querier.yaml
@@ -1,0 +1,24 @@
+# SPDX-License-Identifier: AGPL-3.0-only
+
+# Grants users the ability to query their own audit logs by inheriting the
+# activity.miloapis.com-audit-log-querier Role into iam-user-self-manage.
+#
+# This inheritance ships with the activity service overlay rather than the core
+# control-plane role bundle on purpose: the audit-log-querier Role is provided
+# by the activity stack, so referencing it from the foundational layer would
+# force the core control plane to depend on activity being up first (an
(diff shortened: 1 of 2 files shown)


config/services/activity/roles/iam-user-self-manage-audit-log-querier.yaml
scotwells, line 1: I don't think this would do what you're expecting? This would create a new role called `iam-user-self-manage`, not patch the existing one.

scotwells (changes requested):

PR #678, opened 2026-06-30
docs: add shared operational memory under .claude/memory

ecv:
## Summary

Every Claude Code session on this repo rediscovers the same handful of tooling habits and one piece of milo-specific architecture, because nothing in the tree records them.

This adds a small version-controlled memory next to the agent config, one file per fact plus an index, covering the GitHub CLI habits that keep issue bodies rendering and sub-issues reachable, keeping a local main current, running an unattended loop without permission prompts, and where this repo's ActivityPolicy resources live and how they ship.

Everything environment-specific stayed out, so there are no incident catalogues, cluster names, project IDs, credentials, or deploy-repo paths.

The ActivityPolicy note overlaps the area https://github.com/milo-os/milo/pull/679 touches, so review that entry against it.

## Test plan

- [ ] Each entry holds for this repo on its own and is safe to publish
- [ ] A session that loads the index can act on each note without asking for more context

### DIFF
--- a/.claude/memory/MEMORY.md
+++ b/.claude/memory/MEMORY.md
@@ -1,0 +1,16 @@
+# Shared operational memory
+
+Version-controlled operational notes for working on this repo with Claude Code
+(or any agent). Each entry below points to one file holding one fact. These are
+the *public, repo-portable* subset of operating knowledge — tooling habits, git
+hygiene, unattended-run discipline, and milo-specific architecture that is not
+already obvious from the code or CLAUDE.md.
+
+Keep entries here generic and non-sensitive: no credentials, cluster names,
+project IDs, or internal-only incident detail.
+
+- [gh issue/PR body via --body-file](gh-issue-body-file.md) — use `--body-file` not `--body "$(cat <<EOF)"` so code fences render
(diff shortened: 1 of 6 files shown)


ecv:
Considering this in a public repo. RFC.

kevwilliams (commented):
`activity-policies-owned-by-source-repos.md`'s "known latent bug pattern" doesn't hold up: it claims an unguarded `link(audit.responseObject.metadata.name, audit.objectRef)` recurs across iam/resourcemanager create policies. Checked every `created X` summary in `config/services/activity/policies/{iam,resourcemanager}/*.yaml` (user, group, role, serviceaccount, organization, project) — all already wrap it in `has(audit.responseObject.metadata.name) ? ... : ...` fallback chains, so the DLQ-leak scenario described isn't currently reproducible. (There are unguarded `link(audit.responseObject...)` calls in `userinvitation-policy.yaml`, but for accept/decline/cancel transitions, not the "rejected create" case the note describes.) Publishing this as-is would send a future session chasing a bug that's already fixed. The rest of the paths/commands (activity policies directory, gh body-file, sub-issues GraphQL) check out fine and nothing sensitive is exposed — please just correct or drop that bug-pattern paragraph.

scotwells (approved):

PR #679, opened 2026-06-30
feat(iam): re-home audit-log-querier grant to a standalone activity role (replaces #677)

### LINKED ISSUE #676: Move audit-log-querier role inheritance from core CRD bundle to activity service

ecv:
## Summary

**Replaces #677.** Closes #676.

Delivers the "users can query their own audit logs" capability from the **activity service** overlay *without* mutating the shared `iam-user-self-manage` Role, keeping the foundational milo control plane free of any activity dependency.

## Why #677 doesn't work

#677 added a partial `Role/iam-user-self-manage` manifest (only `spec.inheritedRoles`) to the activity overlay and relied on server-side apply to merge that single entry into the core role.

That assumption fails under the actual GitOps setup (datum-cloud/infra):

- The **core role** is applied by the `milo-core-control-plane-crds` Flux Kustomization (`./crd/overlays/core-control-plane` into `../../../roles`).
- The **partial overlay** would be applied by `milo-activity-policies` (`./services/activity`).
- Flux's `kustomize-controller` server-side-applies **every** Kustomization under a **single shared field manager** (`kustomize-controller`). There is no per-Kustomization manager and no field-manager override on the Kustomization spec.

With one shared manager, each apply re-declares its full field set, so SSA prunes any field that manager owned but no longer lists. The two Kustomizations therefore **prune each other's fields on every reconcile** (`interval: 1h` each). The role flip-flops:

| After reconcile of… | Result |
|---|---|
| `milo-core-control-plane-crds` | `includedPermissions` present, **audit inheritance gone** |
| `milo-activity-policies` | audit inheritance present, **all self-manage permissions gone** |

`inheritedRoles` being `listType=map` only enables clean merge across **distinct** field managers owning distinct keys, which Flux does not provide here. (The companion infra PR's premise that `inheritedRoles` is "an atomic list" is also incorrect, it is map-type, but the shared-field-manager problem stands regardless.)

## What this PR does instead

- **New:** `config/services/activity/roles/activity-self-audit-log-querier.yaml`, a standalone `Role` that inherits `activity.miloapis.com-audit-log-querier`. A distinct object with a single owner means no cross-Kustomization field contention.
- **Edit:** `config/services/activity/kustomization.yaml` wires the role into the activity overlay.
- **Webhook** (`user_webhook.go`): grants the role **per-user, scoped to the user's own `User` resource**, mirroring the existing `user-self-manage` PolicyBinding, so self-audit access stays self-scoped. Gated on a configurable role name (empty means skip), so the core control plane carries **zero activity coupling** when activity is not deployed.
- **Controller** (`user_controller.go`): attaches an owner reference to the binding for GC on user deletion, `IsNotFound`-guarded so it is inert when the activity overlay is absent.

### Why not a static group binding

A single `PolicyBinding` to `system:authenticated` (like `organization-creator-policy`) needs no per-user code, but only `resourceKind: User` is available for a `Group` subject, which would grant **every** user query access to **all** users' audit logs. Rejected as a scope regression.

## Validation

- `go build` on edited packages is clean.
- `task generate:code` produces no manifest diff (no new API types or markers).
- Full builds and tests via CI.

## Coordination

- Part of: datum-cloud/infra#2943, datum-cloud/infra#2939.
- Pairs with: datum-cloud/infra#2953 (its "atomic list" justification should be corrected to "map-type, but shared Flux field manager").

### DIFF
--- a/cmd/milo/controller-manager/controllermanager.go
+++ b/cmd/milo/controller-manager/controllermanager.go
@@ -518,7 +518,7 @@ func Run(ctx context.Context, c *config.CompletedConfig, opts *Options) error {
 				logger.Error(err, "Error setting up organizationmembership webhook")
 				klog.FlushAndExit(klog.ExitFlushTimeout, 1)
 			}
-			if err := iamv1alpha1webhook.SetupUserWebhooksWithManager(ctrl, SystemNamespace, "iam-user-self-manage"); err != nil {
+			if err := iamv1alpha1webhook.SetupUserWebhooksWithManager(ctrl, SystemNamespace, "iam-user-self-manage", "activity-self-audit-log-querier"); err != nil {
 				logger.Error(err, "Error setting up user webhook")
 				klog.FlushAndExit(klog.ExitFlushTimeout, 1)
 			}
--- a/config/services/activity/kustomization.yaml
+++ b/config/services/activity/kustomization.yaml
@@ -6,5 +6,8 @@
 apiVersion: kustomize.config.k8s.io/v1alpha1
 kind: Component
 
+resources:
+  - roles/activity-self-audit-log-querier.yaml
+
 components:
   - policies
--- a/config/services/activity/roles/activity-self-audit-log-querier.yaml
+++ b/config/services/activity/roles/activity-self-audit-log-querier.yaml
@@ -1,0 +1,18 @@
+# SPDX-License-Identifier: AGPL-3.0-only
+
+# Self-scoped audit-log query capability, owned entirely by the activity
+# service overlay. A distinct Role (not a partial patch of iam-user-self-manage),
+# so it is applied by a single field manager and never collides with the core
+# role bundle. Granted per-user, scoped to the user's own User resource, by the
+# user webhook (mirrors the existing user-self-manage PolicyBinding).
+apiVersion: iam.miloapis.com/v1alpha1
+kind: Role
+metadata:
+  name: activity-self-audit-log-querier
+  annotations:
+    kubernetes.io/display-name: Self Audit Log Querier
+    kubernetes.io/description: "Allows a user to query their own audit logs."
+spec:
+  launchStage: Beta
+  inheritedRoles:
+    - name: activity.miloapis.com-audit-log-querier
--- a/internal/controllers/iam/user_controller.go
+++ b/internal/controllers/iam/user_controller.go
@@ -224,6 +224,24 @@ func (r *UserController) ensureOwnerReferences(ctx context.Context, user *iamv1a
 		log.Info("Updated PolicyBinding with owner reference", "user", user.Name)
 	}
 
+	// Update self audit-log PolicyBinding (only present when the activity service
+	// overlay is deployed; otherwise the webhook skips creating it).
+	auditLogPolicyBindingName := fmt.Sprintf("user-self-audit-log-%s", user.Name)
+	auditLogPolicyBinding := &iamv1alpha1.PolicyBinding{}
+	err = r.Client.Get(ctx, types.NamespacedName{Name: auditLogPolicyBindingName, Namespace: "milo-system"}, auditLogPolicyBinding)
+	if apierrors.IsNotFound(err) {
+		// Binding absent: activity overlay not deployed, or webhook hasn't created it yet.
+		log.Info("Self audit-log PolicyBinding not found, skipping", "user", user.Name, "policyBinding", auditLogPolicyBindingName)
+	} else if err != nil {
+		return fmt.Errorf("failed to get self audit-log policy binding: %w", err)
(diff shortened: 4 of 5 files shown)


ecv:
Detail moved out of the description.

Why #677 fails. It added a partial `Role/iam-user-self-manage` (only `spec.inheritedRoles`) to the activity overlay and relied on server-side apply to merge that entry into the core role. In datum-cloud/infra the core role is applied by the `milo-core-control-plane-crds` Flux Kustomization and the partial by `milo-activity-policies`. Flux's kustomize-controller applies every Kustomization under one shared field manager, so each apply prunes the fields the other owns, on every reconcile (`interval: 1h` each):

| After reconcile of | Result |
|---|---|
| `milo-core-control-plane-crds` | `includedPermissions` present, audit inheritance gone |
| `milo-activity-policies` | audit inheritance present, all self-manage permissions gone |

`inheritedRoles` is `listType=map`, which merges cleanly only across distinct field managers, and Flux does not provide them here.

What this change does:

- `config/services/activity/roles/activity-self-audit-log-querier.yaml` adds a standalone Role inheriting `activity.miloapis.com-audit-log-querier`. One object with one owner, so no contention.
- `config/services/activity/kustomization.yaml` wires it into the activity overlay.
- `user_webhook.go` binds each user to that Role, scoped to their own `User` resource, mirroring the `user-self-manage` PolicyBinding. The role name is configurable, and empty skips the binding, so the core control plane carries no activity coupling.
- `user_controller.go` sets an owner reference on the binding for garbage collection on user deletion, guarded by `IsNotFound` so it is inert without the activity overlay.

Rejected alternative. One PolicyBinding to `system:authenticated`, like `organization-creator-policy`, needs no per-user code, but a Group subject only allows `resourceKind: User`, which would let every user query every user's audit logs.

Validation so far: `go build` on the edited packages is clean, and `task generate:code` produces no manifest diff.

Supersedes #677. Part of datum-cloud/infra#2943 and datum-cloud/infra#2939.

scotwells (changes requested):
You shouldn't need to create a separate policy binding for this. It should be granted through role inheritance on the user self-manage role. It should be patched on the infra layer so the system's are de-coupled.

ecv:
better yes

PR #681, opened 2026-07-01
test: migrate project-lifecycle from infra (WIP)

ecv:
> **WIP / DRAFT.** Migration in progress, do not merge yet.

## What

Migrates the **Project lifecycle** Chainsaw test out of `datum-cloud/infra` into milo's local-env harness at `test/resource-management/project-lifecycle/`.

Flow:

1. Create `Organization`, `User`, and `OrganizationMembership`.
2. Create a `Project` in the org control plane and wait for `Ready`.
3. Assert it is listed in both the org and main control planes.
4. Delete it and assert it is gone.

## Why

Per `datum-cloud/infra#3006`, infra's live-env Chainsaw suite is being consolidated into a single "construct" group and single-service API-contract tests are being evacuated to their owning repos, run against milo's **local** apiserver pre-merge. This test validates the `resourcemanager.miloapis.com` Project contract, which milo owns.

The infra original was permanently `skip: true` (blocked on CI SA identity resolution and `User.iam.miloapis.com` uid, infra#2733) and depended on `datumctl` live-auth. Running it here against the local apiserver removes both blockers.

Refs:
- `datum-cloud/infra#3006` (consolidation / evacuation)
- `datum-cloud/infra#3005`

## Adaptation from the infra original

- Dropped `datumctl auth update-kubeconfig` steps and the live-auth `clusters:`/`cluster:` kubeconfig plumbing; targets local `main` and `org` control planes using the sibling `project-creation` kubeconfig convention.
- Removed `spec.skip: true`.
- Added Organization, User, and OrganizationMembership fixtures (the local apiserver requires these before a Project can be created in the org CP).
- Added a `spec.description` block; zero inline comments per repo convention.

`chainsaw lint`: passes.

## Known-uncertain

- Ready-status timing (`Ready=True` within 60s) mirrors the sibling `project-creation` and `project-deletion` tests, but was not executed end-to-end in this branch (no live green CI run performed, do not block on CI here).
- The `assert-project-ready.yaml` status block (`reason: Ready`, `message: Project is ready`) is copied from the sibling assertion. If the local reconciler emits different reason or message text the assert may need loosening.

### DIFF
--- a/test/resource-management/project-lifecycle/01-test-organization.yaml
+++ b/test/resource-management/project-lifecycle/01-test-organization.yaml
@@ -1,0 +1,11 @@
+apiVersion: resourcemanager.miloapis.com/v1alpha1
+kind: Organization
+metadata:
+  name: test-project-lifecycle-org
+  labels:
+    test.miloapis.com/scenario: "project-lifecycle"
+  annotations:
+    kubernetes.io/description: "Organization for testing project lifecycle"
+    kubernetes.io/display-name: "Test Project Lifecycle Organization"
+spec:
+  type: Standard
--- a/test/resource-management/project-lifecycle/02-test-user.yaml
+++ b/test/resource-management/project-lifecycle/02-test-user.yaml
@@ -1,0 +1,10 @@
+apiVersion: iam.miloapis.com/v1alpha1
+kind: User
+metadata:
+  name: user-admin
+  labels:
+    test.miloapis.com/scenario: "project-lifecycle"
+spec:
+  email: admin@test.local
+  givenName: ProjectTest
+  familyName: Admin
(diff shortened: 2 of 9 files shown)


ecv:
Closing. This test duplicates coverage that already exists on `main`.

`test/resource-management/project-deletion` runs the same flow: Organization, User, OrganizationMembership, Project in the org control plane, wait for `Ready`, delete, verify gone from the main control plane. The only step this branch adds is the assert that the Project is listed in the main control plane before deletion.

The branch also drifted from `main`:

- Reuses the cluster-scoped User `user-admin`, which `project-deletion` already owns. Commit `3484fc86` fixed that race between concurrent tests.
- Asserts `status.conditions` as a plain list. `main` moved to `(conditions[?type == 'Ready'])` in `206c4ae2` because the suspension propagator adds a second condition.
- No generated `README.md`.

The one missing assert moves into `project-deletion` in a follow-up PR. infra#3006 asked owner repos for fresh coverage rather than ports, so nothing in infra waits on this branch.

PR #683, opened 2026-07-01
test(e2e): Add ServiceAccount API contract test

ecv:
## Summary

Service accounts had no end-to-end coverage. This test creates one in a project control plane, checks that a new service account starts active, then deletes it and checks it is gone.

Key issuance, authentication, and revocation need the auth service, which the local env lacks, so they stay in infra.

## Test plan

- [ ] The e2e suite passes with the new service account coverage

Related to datum-cloud/infra#3006.

### DIFF
--- a/test/iam/service-account-contract/01-test-organization.yaml
+++ b/test/iam/service-account-contract/01-test-organization.yaml
@@ -1,0 +1,11 @@
+apiVersion: resourcemanager.miloapis.com/v1alpha1
+kind: Organization
+metadata:
+  name: test-service-account-contract-org
+  labels:
+    test.miloapis.com/scenario: "service-account-contract"
+  annotations:
+    kubernetes.io/description: "Organization for testing the ServiceAccount API contract"
+    kubernetes.io/display-name: "Test Service Account Contract Organization"
+spec:
+  type: Standard
--- a/test/iam/service-account-contract/02-test-user.yaml
+++ b/test/iam/service-account-contract/02-test-user.yaml
@@ -1,0 +1,10 @@
+apiVersion: iam.miloapis.com/v1alpha1
+kind: User
+metadata:
+  name: user-admin
+  labels:
+    test.miloapis.com/scenario: "service-account-contract"
+spec:
+  email: admin@test.local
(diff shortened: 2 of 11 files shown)


ecv:
## ⚠️ Action required before this leaves draft

1. **This is a deliberate split, not a full migration.** Only the ServiceAccount **API
   contract** (create → assert `state: Active` default → delete → assert gone) moves here.
   The datumctl key-issuance / auth / revoke / verify-rejected roundtrip from the original
   infra test **stays in infra** as a construct test (it needs the auth provider / Zitadel,
   absent in milo local env). infra#3006's disposition note is being updated to reflect this.
2. **`Ready` / `status.email` are not settable locally** — they come from the auth-provider
   integration; no in-tree milo controller sets them. Hence the test asserts create/default/
   delete only. Confirm that's the intended contract scope for the local suite.
3. **Possible OpenFGA gate.** If ServiceAccount CREATE is authorization-gated locally, this
   test needs the `requires: authorization-provider` selector label (the suite runs with
   `--selector "requires!=authorization-provider"`). Verify against a running local apiserver.

Not run end-to-end yet (only `chainsaw lint`). Tracks datum-cloud/infra#3006.

ecv:
Resolving the three items from the July draft note.

1. **Split scope.** Matches the revised split note on datum-cloud/infra#3006. The key issuance, authentication, and revocation roundtrip stays in infra. Only the API contract lives here.
2. **`Ready` and `status.email` not settable locally.** Confirmed. The test asserts create, default state, delete, and absence at the platform root. Nothing in the local suite can assert the auth-provider fields.
3. **OpenFGA gate.** Not gated. CI run [34055194481](https://github.com/milo-os/milo/actions/runs/34055194481) created the ServiceAccount and passed without the `requires: authorization-provider` selector.

One finding came out of the first green run. A ServiceAccount applied with no `spec` key stores no `spec.state`, so the fixture sets `spec: {}` for now. #773 tracks the CRD default fix.

The full e2e suite passed on `fd294ff5`. CI on the current head is running.

scotwells (approved):

PR #712, opened 2026-07-09
fix: self-delete OrganizationMembership when its Organization is gone

0xmc:
## Summary

`OrganizationMembershipController` already self-deletes a membership once
its `User` is gone (`UserNotFoundReason`), via a two-pass check: the first
reconcile records the failure reason, the second reconcile - triggered by
that status write re-enqueuing the object - sees the same reason still set
and deletes for real.

There was no equivalent for the referenced `Organization` being gone
(`OrganizationNotFoundReason`). Deleting an Organization left its
memberships permanently stuck reporting `OrganizationNotFound`, and since
nothing ever deleted the membership, the owner `PolicyBinding` it manages
(owned by the membership, not the Organization) was never garbage
collected either.

This showed up as part of a live staging incident: `PolicyBindingsNotReady`
and `OrganizationMembershipsNotReady` were both firing, with several
memberships/bindings that had been silently orphaned for months.

- Add the same two-pass self-delete for `OrganizationNotFoundReason`,
  mirroring the existing `UserNotFoundReason` path.
- Add `Reconcile`-level test coverage for **both** self-delete paths -
  neither had any test coverage before this change.

## Test plan

- [x] `go test ./internal/controllers/resourcemanager/...` - new and
  existing tests pass
- [x] `go build ./...` - builds clean
- [x] `gofmt -l` - clean

Related to a companion fix for a second, unrelated orphaned-`PolicyBinding`
class (project-owner bindings surviving their creating `User`'s deletion) -
coming as a separate PR since the two bugs share no files.

### DIFF
--- a/internal/controllers/resourcemanager/organization_membership_controller.go
+++ b/internal/controllers/resourcemanager/organization_membership_controller.go
@@ -104,6 +104,26 @@ func (r *OrganizationMembershipController) Reconcile(ctx context.Context, req ct
 	if err := r.Client.Get(ctx, organizationKey, &organization); err != nil {
 		if apierrors.IsNotFound(err) {
 			logger.Info("referenced organization not found", "organization", organizationMembership.Spec.OrganizationRef.Name)
+
+			// Two-pass self-delete: if we already set OrganizationNotFound on a
+			// previous reconcile, delete the membership now so it does not linger
+			// after the owning Organization is gone. The second reconcile is
+			// triggered by the status condition update below, which re-enqueues
+			// the membership. This mirrors the UserNotFound self-delete below;
+			// without it, memberships (and the PolicyBindings they own) are
+			// orphaned forever once their Organization is deleted, since nothing
+			// else garbage collects them.
+			existingCondition := apimeta.FindStatusCondition(organizationMembership.Status.Conditions, OrganizationMembershipReady)
+			if existingCondition != nil && existingCondition.Reason == OrganizationNotFoundReason {
+				logger.Info("deleting OrganizationMembership because referenced organization no longer exists",
+					"membership", organizationMembership.Name,
+					"organization", organizationMembership.Spec.OrganizationRef.Name)
+				if err := r.Client.Delete(ctx, &organizationMembership); err != nil && !apierrors.IsNotFound(err) {
+					return ctrl.Result{}, fmt.Errorf("failed to self-delete organization membership: %w", err)
+				}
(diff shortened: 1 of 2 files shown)


scotwells:
Does this interfere with the webhook that automatically created the organization membership for the organization when the org is created? That would result in the org membership being created before the organization is fully created.

mattdjenkinson (approved):

0xmc:
@scotwells as I understand it, the OrganizationMembership can't exist until OrganizationController has already confirmed the Organization exists.

scotwells:
@0xmc the organization membership can exist, but the controller won't do anything with it until the organization it's referencing exists. This is pretty typical with k8s controllers when resources referencing other resources may be created in a different order.

With organizations, we have a webhook setup that intercepts the request to create an organization that creates the namespace and adds the authenticated user as the owner of the organization. Once both of those resources have been created in the system, the Organization is created. 

https://github.com/milo-os/milo/blob/1ef21a2b66d62d127974543035550cf7264b81d7/internal/webhooks/resourcemanager/v1alpha1/organization_webhook.go#L133-L136

https://github.com/milo-os/milo/blob/1ef21a2b66d62d127974543035550cf7264b81d7/internal/webhooks/resourcemanager/v1alpha1/organization_webhook.go#L155-L165

0xmc:
@scotwells got it, thanks.  But back to the original question, this change is enabling deletion of an OrganizationMembership after its Organization has been deleted.  But like you wrote, an OrganizationMembership won't be touched until the Organization exists.  So I don't see how this change might effect the webhook that's creating the Organization.

scotwells:
@0xmc the Organization isn't actually created in the system until the OrganizationMembership is created. The webhook intercepts the request before it's stored in the storage layer. So the controllers will see the OrganizaitonMembership first and then will see that the Organization for it doesn't exist.

0xmc:
@scotwells ack.  working on a fix modeled after the grace period in https://github.com/milo-os/milo/blob/main/internal/quota/controllers/lifecycle/ownership.go

scotwells:
@0xmc do we know exactly what sequence of events led up to this and caused the membership to be orphaned? I'd expect the organization deletion to delete the entire namespace which should delete the memberships of the org. Audit logs would be able to piece together the timeline.

scotwells (approved):

scotwells:
Dug into this and confirmed the organization's namespace was never stamped with the owner reference because it was created / deleted so quickly.

Issue #684, opened 2026-07-02
Webhook server only starts on the elected leader, breaking multi-replica deployments


ecv:
## Summary

The controller-manager appears to start its webhook server (`:9443`) **only on the elected leader**. In a multi-replica deployment, non-leader replicas stay `Ready` (their health/API port `:6443` serves fine and passes probes) and therefore remain in the Service endpoints — but their `:9443` webhook listener is never opened.

A `Service` fronting `milo-controller-manager:9443` then load-balances webhook admission calls across all replicas. Every request routed to a non-leader hits a closed listener and fails:

```
failed calling webhook "mpolicybinding.iam.miloapis.com":
Post "https://milo-controller-manager.<ns>.svc:9443/mutate-...":
dial tcp <clusterIP>:9443: connect: connection refused
```

With N replicas, roughly `(N-1)/N` of webhook calls fail. With `failurePolicy: Fail` this breaks admission for the affected resources intermittently.

## Impact observed

In a 2-replica production deployment this manifested as ~50% of `PolicyBinding` admission dry-runs failing (`mpolicybinding.iam.miloapis.com`), which stalled GitOps reconciliation of the control-plane CRD/RBAC bundle and cascaded to dependent resources. Workaround was to drop to a single replica (leader == the only pod == the only `:9443` listener), which restores 100% webhook success but removes HA/failover.

## Expected behavior

The webhook server should serve on **all** replicas, independent of leader election — only the reconcile/controller loops should be leader-gated. This is the controller-runtime default: the webhook server is registered as a non-leader-election runnable so every replica can answer admission requests, and the Service can safely load-balance across them.

## Suggested fixes (either/both)

1. Register the webhook server as a **non-leader-election runnable** so it starts on every replica regardless of leadership. (Preferred — matches controller-runtime defaults.)
2. Add a **readiness gate on `:9443`** so a replica not serving the webhook is pulled from Service endpoints. (Mitigation; still leaves only one serving backend if webhook stays leader-gated.)

## Blocking

Until this is fixed, multi-replica controller-manager is unsafe — we are pinned to a single replica downstream.

ecv:
This should clear in https://github.com/milo-os/milo/pull/675

ecv:
[...]

PR #717, opened 2026-07-14
fix: run controller-manager with 2 replicas for webhook HA

ecv:
## What

Run `milo-controller-manager` with **2 replicas** so admission webhooks stay available across a pod outage.

Implements datum-cloud/infra#2365 ("Run milo-controller-manager in HA mode (2 replicas)").

## Why

Admission webhooks already start on **every** replica via the dedicated non-leader-election webhook manager (`startCoreControlPlaneWebhooks`). They serve before and independent of leader election. The prerequisite fix for that (webhook only started on the elected leader) landed in `main` via #9ffd87f, #c179dbf, and #08c4dae (tracked as #684). A separate webhook Deployment is **not** needed to decouple webhook availability from leadership.

The one remaining gap behind the bootstrap boot loop (datum-cloud/infra#1526, milo#716) was that the controller-manager ran a **single** replica (`replicas: 1`), so admission still died whenever that pod did (crashloop, OOM #631, rollout). With one replica the all-replica webhook code has no other replica to fall back to.

## How

- `config/controller-manager/base/deployment.yaml`: `replicas: 1` becomes `2`.
- Document the invariant at `startCoreControlPlaneWebhooks` so future work doesn't reintroduce a separate webhook Deployment to solve availability.

## Notes

- Multi-replica is safe now: the webhook server listens on all replicas (#684), not just the leader; reconcile loops remain leader-gated.
- milo#389 (leader election hardcodes `datum-system`) is a cross-namespace collision bug, not a same-namespace replica-count blocker. The deployment already passes `--leader-elect-resource-namespace=milo-system`.
- Supersedes the separate-webhook-Deployment approach earlier in this PR (per review feedback); milo#716 closed in favour of infra#2365.

### DIFF
--- a/cmd/milo/controller-manager/webhooks.go
+++ b/cmd/milo/controller-manager/webhooks.go
@@ -160,6 +160,12 @@ func newWebhookRESTMapper(ctx context.Context, rootClientBuilder clientbuilder.C
 // startCoreControlPlaneWebhooks starts the admission webhook server on every replica.
 // It is invoked before leader election when leader election is enabled so followers
 // also listen on the webhook port and Service endpoints remain healthy.
+//
+// This is the dedicated, non-leader-election webhook manager: admission stays
+// available even when leadership has not been (or cannot be) acquired, so it is
+// independent of the leader-gated reconcile loops. Do NOT add a separate webhook
+// Deployment to "decouple" it — that is already done here. To harden webhook
+// availability, run the controller-manager with >=2 replicas.
 func startCoreControlPlaneWebhooks(
 	ctx context.Context,
 	opts *Options,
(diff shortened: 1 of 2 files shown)


scotwells:
@ecv we shouldn't need to run it in a separate deployment, we just need to make sure that the webhook always starts even when leadership election hasn't succeeded. This is pretty standard with k8s controllers / webhooks.

ecv:
iirc this is blocked for some Good Reason

ecv:
My shell thinks we landed the stuff blocking this from going multireplica. https://github.com/milo-os/milo/pull/675 specifically

ecv:
This moves the bundle default to two replicas, matching what Datum pins for its own deployment in https://github.com/datum-cloud/infra/pull/4404.

The two are independent. Datum overrides the replica count on the rendered bundle, so this change does not alter what Datum runs, and that change does not need this one. What this does is stop the shipped default from contradicting the recommendation in the comment above `startCoreControlPlaneWebhooks`, which matters for anyone deploying the bundle without an override.

Holding as draft until the Datum change reaches trunk. The default should not move ahead of the deployment that validates it, so that the first consumer to pick up two replicas is not also the first to try it.

PR #722, opened 2026-07-15
feat(resourcemanager): implement ProjectSuspension CRD, validation webhooks, and IAM roles

JoseSzycho:
## Overview
Introduces the `ProjectSuspension` platform primitive to allow reversibly suspending and reinstating projects, along with its validation webhooks, IAM protected resource mappings, and dedicated RBAC roles.

## Key Changes
- **API & CRD**: Registered the `ProjectSuspension` schema, setting validation parameters and CEL immutability rules.
- **Webhooks**: 
  - Implemented mutating/validating webhooks to manage suspension requests.
  - Enforces `reinstateAuthority` on deletion requests, blocking non-operators from reinstating suspensions when restricted, while permitting system garbage collection when the parent project is deleted.
- **IAM & RBAC**:
  - Registered `ProjectSuspension` as a `ProtectedResource`.
  - Created new dedicated staff roles (`resourcemanager-projectsuspensions-reader/editor/admin`).
  - Added read permissions to the existing `project-viewer` role.
- **Testing**: Added unit tests for webhook admission rules and integrated end-to-end scenarios using Chainsaw.

Main enhancement:

Issue:
- https://github.com/milo-os/milo/issues/718


config/roles/project-viewer.yaml
```diff
@@ -10,4 +10,7 @@ spec:
   includedPermissions:
     - resourcemanager.miloapis.com/projects.get
     - resourcemanager.miloapis.com/projects.list
-    - resourcemanager.miloapis.com/projects.watch
\ No newline at end of file
+    - resourcemanager.miloapis.com/projects.watch
+    - resourcemanager.miloapis.com/projectsuspensions.get
+    - resourcemanager.miloapis.com/projectsuspensions.list
+    - resourcemanager.miloapis.com/projectsuspensions.watch
```
scotwells: @JoseSzycho we don't want consumers to be able to view these resources. Only platform operators should be able to view them.

internal/webhooks/resourcemanager/v1alpha1/projectsuspension_webhook.go
```diff
@@ -0,0 +1,148 @@
...
+
+	if obj.Spec.ReinstateAuthority == resourcemanagerv1alpha1.AuthorityOperator {
+		if !isOperator {
+			return nil, errors.NewForbidden(
+				resourcemanagerv1alpha1.GroupVersion.WithResource("projectsuspensions").GroupResource(),
+				obj.Name,
+				fmt.Errorf("only operators can lift/delete this suspension"),
+			)
+		}
+	}
```
scotwells: I don't think we need this check in here since the `delete` access would determine whether the user has the ability to delete a project's suspension. 

The authority is more of a signal on who needs to take action to get the project re-instated. In the case of the consumer they may need to pay a bill to have the system re-instate their project or if it's an operator, the consumer needs to reach out to the operator to get their project re-instated.

scotwells (changes requested):
Overall looks good. Just some minor tweaks.

config/roles/project-viewer.yaml
JoseSzycho: Thanks for the review. I removed the permissions.

I added them, as I thought that the users should be able to know why their project got suspended. 

When I write the controllers, I will add further information in the Project status so they can get the information and understand whats going on.

internal/webhooks/resourcemanager/v1alpha1/projectsuspension_webhook.go
JoseSzycho: That makes sense. 

Thank you for the explanation, I just removed these checks.

config/roles/project-viewer.yaml
scotwells: @JoseSzycho we want to avoid surfacing too much information to users, especially information that should be internal only.

scotwells (approved):

PR #728, opened 2026-07-17
feat(resourcemanager): implement project suspension propagation and lifecycle management

JoseSzycho:
This PR implements the project suspension reconciler and propagation system. It tracks active `ProjectSuspension` resources targeting a `Project` and propagates the aggregated suspension state (reasons and references) onto the `Project`'s status and conditions. When no suspensions remain active, the project is returned to active status.

### 1. API Changes (`pkg/apis/resourcemanager/v1alpha1`)
- Added structural fields to `ProjectStatus` to track a slice of `ProjectSuspensionInfo` structs. Each suspension detail retains the name (`ProjectSuspensionReference`) and the suspension reason (`ProjectSuspensionReason`).
- Added the `+kubebuilder:selectablefield:JSONPath=".spec.projectRef.name"` annotation to `ProjectSuspension` to enable efficient field indexing.
- Generated updated manifests and deepcopy methods (`task generate`).

### 2. Controller Implementation (`internal/controllers/resourcemanager`)
- Implemented `ProjectSuspensionPropagatorController` (`projectsuspension_propagator_controller.go`) which reconciles `Project` resources.
- Registers a field indexer on `spec.projectRef.name` for fast O(1) lookups of suspensions.
- Merges all active suspensions into `status.suspensions`, sorted deterministically by name.
- Sets the `Suspended` status condition to `True` (reason `Suspended`) if there are active suspensions, or `False` (reason `Active`) if none are present.
- Updates the status using standard `Status().Patch` with the field owner `"project-suspension-propagator-controller"` and semantic diff checks to prevent conflicting writes.
- Registered the controller with the Manager in `controllermanager.go`.

### 3. Testing
- **Unit Tests**: Implemented test scenarios in `projectsuspension_propagator_controller_test.go` covering active, lifted, and absent suspensions.
- **E2E Tests**: Created a new E2E integration test suite under `test/resource-management/project-suspension-lifecycle` using Chainsaw. The test asserts the step-by-step lifecycle of adding multiple suspensions, verifying the status updates correctly, and removing them one by one until the project transitions back to active status.
- **E2E Assertions**: Updated all project E2E assertion files across the repository to filter the `Ready` and `Suspended` conditions using parenthesized JSONPath selectors, isolating checks from slice length mismatches.

Related to:
- https://github.com/milo-os/milo/issues/718

Main enhancement:


JoseSzycho:
I would appreciate at least a quick review on this new Project Status field:

https://github.com/milo-os/milo/blob/206c4ae21778f22ba051cb5c934141a12479d65a/pkg/apis/resourcemanager/v1alpha1/project_types.go#L23-L31

As I need this information to update some CR in the service catalog.

scotwells:
@JoseSzycho I think the reason / suspension timestamp / reinstate authority would be enough? What's the use-case for having the full reference in there?

JoseSzycho:
@scotwells I added the full reference as usually is easier for the front team to have these references in order to fetch the necessary resources on the Staff portal.

Would you prefer me to remove it, and add the suspension timestamp and the reinstate authority?

scotwells:
@JoseSzycho for this view on it yeah. The portal could just resolve it through a field selector on the project name since that's embedded in the project suspension resource.

JoseSzycho:
@scotwells thanks for this really quick review. I just made this changes.

config/crd/bases/resourcemanager/resourcemanager.miloapis.com_projects.yaml
```diff
@@ -135,6 +135,38 @@ spec:
...
+                    affecting the project.
+                  properties:
+                    reason:
+                      description: Reason is the category of suspension.
+                      enum:
+                      - Abuse
+                      - Billing
+                      - Compliance
+                      - Administrative
+                      type: string
```
scotwells, line 153: Can we get "Fraud" added to this list? It's separate from Abuse.

scotwells (dismissed):
Looks good. One note for follow up.

config/crd/bases/resourcemanager/resourcemanager.miloapis.com_projects.yaml
JoseSzycho: Sure, why not! ?

Commit pushed :)

scotwells (approved):

PR #731, opened 2026-07-21
feat(admission): implement project suspension admission plugin and e2e test scenarios

### LINKED ISSUE #719: Block new work in a suspended project (admission enforcement)
While a project is suspended, its control plane rejects writes (reads still allowed) with a clear, typed reason.

## Acceptance criteria
- [ ] **Evaluate `ValidatingAdmissionPolicy` (CEL) first** — determine whether it can block writes without a bespoke webhook, and confirm the suspension signal is reachable by a CEL policy in the *project* control plane (the authoritative `Suspended` condition lives on the `Project` in the *core* control plane). Record the finding.
- [ ] Implement via `ValidatingAdmissionPolicy` if viable, otherwise a webhook.
- [ ] If the signal is not reachable, coordinate with the primitive work to project a minimal marker into the project control plane for the policy to key off.
- [ ] Deny responses carry a typed reason so clients can surface "this project is suspended."

**Dependencies:** platform primitive (suspension signal).

---
Part of the **Project Suspension & Lifecycle Controls** enhancement (datum-cloud/enhancements#800). Rendered proposal: https://github.com/datum-cloud/enhancements/blob/enhancement/project-suspension/enhancements/platform/project-suspension/README.md

JoseSzycho:
This PR implements the `ProjectSuspensionEnforcement` admission plugin in the Milo API server to enforce project suspension policies. When a `Project` resource is marked as suspended (via an active `ProjectSuspension`), the admission plugin intercepts mutation requests (`CREATE`, `UPDATE`) targeting that project's control plane and rejects them with a structured `403 Forbidden` status (`ProjectSuspended` cause in `Details.Causes`).

### Key Changes

- **Enforcement**: Intercepts resource creation and modification in suspended projects, returning a `403 Forbidden` error with `ProjectSuspended` details so clients can distinguish suspension from standard RBAC denials.

- **Exemptions**:
  - **Superusers (`system:masters`)**: Unconditionally exempted so background cluster controllers and system services can perform administrative tasks (e.g., scaling down workloads, pausing services).
  - **Read-only & Delete operations**: `GET`, `LIST`, `WATCH`, and `DELETE` operations bypass suspension enforcement.
  - **Mid-Deletion Resources (`isAlreadyTerminating`)**: `UPDATE`/`PATCH` operations on objects that already have a `deletionTimestamp` (such as finalizer removals) are allowed. This ensures resources mid-deletion can complete teardown without getting stuck in `Terminating` state.
  - **Access Reviews**: `subjectaccessreviews` and `localsubjectaccessreviews` bypass suspension checks.


internal/apiserver/admission/plugin/projectsuspension/admission.go
```diff
@@ -0,0 +1,323 @@
...
+		p.cache.Add(projectID, (*suspensionState)(nil), p.cacheTTL)
+		return nil, nil
+	case err != nil:
+		// Fail open: suspension enforcement is a policy safety net, not a
+		// security boundary. Don't let an unrelated infrastructure error
+		// (root apiserver reachability) block all writes to the project.
+		p.logger.V(2).Info("failed to get project for suspension check, failing open",
+			"projectID", projectID, "error", err)
+		return nil, nil
+	}
```
scotwells: I'd use an informer and maintain a cache of suspended projects instead of resolving it at request time. Otherwise the in-memory cache of project suspension state is going to get out of sync as projects get unsuspended.

scotwells (changes requested):
Good use of the admission plugin here. I like the control it gives us around error messaging. My only feedback on the implementation is we need to use an informer to maintain the cache of suspended projects.

Did you happen to look at what the experience using [an Admission Policy](https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy) would provide? Hypothetically we could configure an Admission Policy when as project is suspended to prevent changes made by consumers.

JoseSzycho:
@scotwells I investigated the use of an Admission Policy, how ever I wanted a little more of flexibility and a plugin approach gave me this little extra flexibility I needed. As I'm sure we would want to make some tweaks in the future, and with the plugin this is going to be easy to make.

I'd like to clarify that I didn't even try to make it work with the Admission Policy, as I discarded that idea before giving it a chance because of the lack of flexibility.

internal/apiserver/admission/plugin/projectsuspension/admission.go
JoseSzycho: you're 100% right. I'm in great favor of using informers for creating cache, It completely forgot to add one.

I already added it an e2e tests are passing.

internal/apiserver/admission/plugin/projectsuspension/cache.go
```diff
@@ -0,0 +1,112 @@
...
+	switch {
+	case apierrors.IsNotFound(err):
+		return nil, nil
+	case err != nil:
+		c.logger.V(2).Info("failed to get project for suspension check, failing open",
+			"projectID", projectID, "error", err)
+		return nil, nil
+	}
+
+	return parseSuspensionState(obj), nil
```
scotwells: We shouldn't fallback to in-line requests. What you'd want to do instead of configure the `/readyz` endpoint to return not-ready until the informer reports being in-sync. You can refer to the quota admission plugin for an example of this being done today: 

https://github.com/milo-os/milo/blob/c6db1900350b7ecaffe152f9f4486d4fd26da866/cmd/milo/apiserver/config.go#L457-L458

https://github.com/milo-os/milo/blob/c6db1900350b7ecaffe152f9f4486d4fd26da866/pkg/quota/admission/register.go#L38-L65

internal/apiserver/admission/plugin/projectsuspension/cache.go
```diff
@@ -0,0 +1,112 @@
...
+// If the item exists in the informer cache, it performs a fast in-memory store lookup.
+// If the item is absent and the cache isSynced, it returns nil (not suspended).
+// If the item is absent and the cache is not yet synced, it falls back to a live Get.
+func (c *ProjectSuspensionCache) GetSuspensionState(ctx context.Context, projectID string) (*suspensionState, error) {
+	item, exists, err := c.informer.GetStore().GetByKey(projectID)
+	if err == nil && exists {
+		if unstr, ok := item.(*unstructured.Unstructured); ok {
+			return parseSuspensionState(unstr), nil
+		}
+	}
```
scotwells: I'd recommend having the informer maintain a concurrent safe map that only has suspended projects in it so you can quickly determine whether the project is suspended instead of having to re-parse a project's entire response body. I would also see if there's a way you can configure the informer so you don't need to maintain the entire list of projects in memory, since you can maintain a separate map.

scotwells (changes requested):

internal/apiserver/admission/plugin/projectsuspension/cache.go
JoseSzycho: thanks for the feedback. 

It was quite difficult to achieve this, but was doable with the newer go runtime we use

internal/apiserver/admission/plugin/projectsuspension/cache.go
```diff
@@ -17,60 +18,169 @@ import (
...
+
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if state == nil || !state.Suspended {
+		delete(s.suspended, u.GetName())
+		return nil
+	}
+	s.suspended[u.GetName()] = state
+	return nil
+}
```
scotwells, line 115: Something we should test for is whether an Update call is seen after a finalizer is removed. I would hope the informer was smart enough to register this as a Delete instead of an Update.

internal/apiserver/admission/plugin/projectsuspension/cache.go
JoseSzycho: Thanks ! I've learned something new :) 

Changes are complete.

test/admission/projectsuspension/chainsaw-test.yaml
scotwells, line 1: @JoseSzycho we should make sure this test also goes through the process of unsuspending the project and confirm that the user can modify resources in the project again.

internal/apiserver/admission/plugin/projectsuspension/cache.go
JoseSzycho: the finalizer removal is seen as an Update event. Once the last finalizer is gone, etcd deletes the record and K8s fires the Delete event.

Asked Claude to add the TestInformerCache_TerminatingProjectAndFinalizerRemoval to cover this exact sequence.

test/admission/projectsuspension/chainsaw-test.yaml
JoseSzycho: Sure, added test for verify thins. 

This was previously validated in test/resource-management/project-suspension-lifecycle, but I think it have more send to make it in the `admission` tests.

internal/apiserver/admission/plugin/projectsuspension/admission.go
```diff
@@ -0,0 +1,293 @@
...
+	reasons := uniqueSortedReasons(s.Suspensions)
+	reasonList := strings.Join(reasons, ", ")
+	if reasonList == "" {
+		reasonList = "unspecified"
+	}
+
+	msg := fmt.Sprintf(
+		"project %q is suspended (%s) and cannot accept new writes while suspended",
+		projectID, reasonList,
+	)
```
scotwells, line 266: "writes" is a little technical. Maybe we should make this more user-friendly and go with "updates"?

scotwells (approved):
Overall looks good! One minor comment on the error messaging used.

internal/apiserver/admission/plugin/projectsuspension/admission.go
JoseSzycho: of course! Created a PR for that https://github.com/milo-os/milo/pull/733

Issue #190, opened 2025-07-16
Organization’s namespace not being deleted when the org is deleted.


mustela:
It seems that when you try to delete an organization, the namespace is not being deleted. As a result, when retrieving a list of organizations, the deleted ones still appear, along with the following error in the response:

```
{
  "lastTransitionTime": "2025-07-08T08:00:06Z",
  "message": "Organization 'abcd-zks0vd' does not exist. Please ensure the organization name is correct and the organization has been created.",
  "observedGeneration": 1,
  "reason": "OrganizationNotFound",
  "status": "False",
  "type": "Ready"
}
```

scotwells:
The owner references are in place now so the namespace will be garbage collected when the organization is deleted. 

```yaml
▶ kubectl -n datum-system exec pods/kubectl-debug -it -- kubectl --kubeconfig /etc/kubernetes/config/kubeconfig get namespaces organization-personal-org-be933431 -o yaml
apiVersion: v1
kind: Namespace
metadata:
  creationTimestamp: "2025-07-03T21:29:37Z"
  labels:
    kubernetes.io/metadata.name: organization-personal-org-be933431
    resourcemanager.miloapis.com/organization: personal-org-be933431
    resourcemanager.miloapis.com/type: organization
  name: organization-personal-org-be933431
  ownerReferences:
  - apiVersion: resourcemanager.miloapis.com/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: Organization
    name: personal-org-be933431
    uid: 3f1ad1d3-c348-4da4-a2c9-95f4819ac9e5
  resourceVersion: "11568448"
  uid: f7e6b8d7-660f-4b53-8874-38d024483de1
spec:
  finalizers:
  - kubernetes
status:
```

PR #734, opened 2026-07-22
fix: garbage collect organization namespace on Organization deletion

scotwells:
## Summary

Deleting an Organization didn't reliably delete its namespace, leaving `OrganizationMembership` and `PolicyBinding` resources orphaned inside it indefinitely. Root cause: the namespace was created without an owner reference to its Organization, so Kubernetes had nothing to cascade-delete from. A background reconcile was supposed to patch that reference in after the fact, but a fast create-then-delete (seen in staging, and common in e2e tests) can delete the Organization before that reconcile runs.

Fix: set the owner reference at namespace creation time instead of relying on a later reconcile. This closes the race entirely.

Regression in the fix for #190: that issue added the owner reference, but only via an async reconcile, which is the race this PR closes.

## Test plan

- [x] New unit test asserting the namespace gets the owner reference, and that an existing namespace is left alone
- [x] New chainsaw e2e test (`test/resource-management/organization-fast-deletion`): creates an Organization, deletes it with no intervening wait, and confirms the namespace is still garbage-collected
- [x] `go build`, `go test`, `go vet`, `gofmt` all clean


0xmc:
The change looks good to me, we'll just need to make sure any existing orphaned are cleaned up.

@scotwells do you want to update the branch or should I before approval?

kevwilliams:
Dug into the `test-environment-validation` failures on this branch — they're not test flakiness, they're a real race this change introduces in the control plane itself.

**Sequence, reproduced locally with controller/apiserver logs streamed live** (all within ~16ms, for org `st-test-org`):

1. `10:53:16.011` — Inside `OrganizationValidator.ValidateCreate` (a *validating* admission webhook on the Organization CREATE), `EnsureOrganizationNamespace` fires a nested `POST /api/v1/namespaces` → 201. The namespace is created with an owner reference to the Organization.
2. `10:53:16.012` — The garbage collector, independently watching everything, sees the new namespace's owner reference and does a live verification `GET /apis/resourcemanager.miloapis.com/v1alpha1/organizations/st-test-org` → **404**.
3. `10:53:16.016` — *Only now* does the outer `POST .../organizations` actually return `201 Created`.
4. `10:53:16.021` — GC, having concluded the owner doesn't exist, deletes the namespace it just pointed at: `DELETE /api/v1/namespaces/organization-st-test-org` → 200.

Root cause: Kubernetes runs validating admission webhooks *before* the object is persisted to the backing store (confirmed against `k8s.io/apiserver/pkg/registry/generic/registry/store.go`: `createValidation` runs before the actual write). So when `ValidateCreate` synchronously creates a namespace owned by "the Organization currently being created," that Organization doesn't exist yet as far as any other reader — including the garbage collector — is concerned. If GC's own concurrent processing happens to verify ownership inside that window (far more likely under a parallel test suite hammering the API, but possible anytime), it sees a dangling reference and cascade-deletes the namespace within milliseconds of it being created — well before the org's own create call has even returned.

This lines up with everything observed in CI: the failures are non-deterministic (pure race against GC's queue), unrelated to any prior namespace state, hit brand-new orgs with unique names, and are **new** — before this PR, namespaces had no owner reference at creation time, so GC had no grounds to touch them. It reproduced locally in a fresh KinD cluster on a full `task test:end-to-end` run, hitting a different random subset of org-creating tests each time (`quota-enforcement-edge-cases`, `quota-core-functionality`, `quota-grant-creation-policy`, `structured-type-enforcement`, `project-deletion` in one run; `organization-membership-owner-protection`, `quota-manual-claim-preservation`, `organization-membership-role-assignment` in others).

I pushed a test-side patch (9459a2f) that adds/hardens leftover-namespace cleanup in three of the originally-failing tests — that's still useful hygiene, but it doesn't address this race, since it's a control-plane timing issue, not a test-ordering one. Any org-creating test can still hit it.

Planning to follow up with a fix that avoids depending on GC verifying an owner mid-creation — most likely moving the owner-reference-set into the first post-commit reconcile (which fires immediately via watch, well after the object exists) instead of the pre-commit validating-webhook path, while keeping a finalizer on the Organization to close the original fast-delete-orphan window this PR was written to fix. Will push that as a follow-up commit here.

kevwilliams:
Pushed the actual fix (4c014e1).

`EnsureOrganizationNamespace` no longer sets the owner reference synchronously — that's what raced against the garbage collector. `OrganizationController.Reconcile` now sets it once the Organization is genuinely committed and watchable, and `OrganizationMutator.Default` adds an `organizationNamespaceFinalizer` synchronously at admission (finalizers don't need the cross-object existence check an owner reference does, so this is race-free) to close the original fast-create-then-delete window without depending on GC timing.

Verified locally with a full `task test:end-to-end` run, controller/apiserver logs streamed throughout: 30 passed, 0 failed, 3 skipped — no `timed out waiting for the condition on namespaces` anywhere, including on a cluster that still had state left over from a prior failing run. CI should confirm on the next push.

mattdjenkinson (approved):

PR #741, opened 2026-07-30
feat: initialize dedicated multicluster manager for note webhooks in controller-manager

JoseSzycho:
# fix: register Notes webhooks on all controller-manager replicas

## Summary

Creating or editing a Note or ClusterNote in staging failed admission with: `failed calling webhook "mnote.notes.miloapis.com": failed to call webhook: the server could not find the requested resource`.

When leader election is enabled on the core control plane, `milo-controller-manager` starts its admission webhook server on every replica (before leader election) so the webhook Service stays healthy, then registers controller-only logic on whichever replica wins leadership. That split had a gap: the pre-election path registered every core webhook except Notes/ClusterNotes (they need a multicluster manager to resolve a Note's subject when it lives in a project cluster, which wasn't available yet), and the leader-only path only registered them when webhooks were *not* already running on all replicas. Combined, the two conditions meant Notes/ClusterNotes webhooks were never registered anywhere, so every request to their endpoints hit a 404.

The fix builds a dedicated multicluster manager inside `startCoreControlPlaneWebhooks`, following the same provider/`EngageAlways`/local-cluster-engage pattern the quota system already uses, and registers the full webhook set (including Notes) against it. This runs on every replica, matching the rest of the core webhooks.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./cmd/milo/controller-manager/...`
- [x] `gofmt -l` reports no formatting issues
- [x] Deploy to staging and confirm creating/editing a Note or ClusterNote no longer 404s on any replica (not just the leader)

--

Related to:
- https://datum-inc.slack.com/archives/C0AKA02CNKT/p1785423206578239

### DIFF
--- a/cmd/milo/controller-manager/webhooks.go
+++ b/cmd/milo/controller-manager/webhooks.go
@@ -20,6 +20,7 @@ import (
 	"k8s.io/klog/v2"
 	controllerruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/cluster"
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 	mcmanager "sigs.k8s.io/multicluster-runtime/pkg/manager"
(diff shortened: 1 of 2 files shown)


ecv (approved):

ecv:
what issue did this fix? (looking for a hyperlink to a github issue, i can read the description)

PR #743, opened 2026-08-03
feat: add website field to org contact info

mattdjenkinson:
## Summary

Organizations have no field to record a website in their contact info. This adds an optional `website` field to `OrganizationContactInfo`. It accepts a bare domain (`example.com`) or a full URL; no format validation is enforced, since a bare domain is a valid value on its own.

Verified `go test ./internal/webhooks/resourcemanager/...` and `./internal/controllers/resourcemanager/...` pass, and regenerated the CRD manifest and API docs via `task generate:code` and `task generate:docs`.


JoseSzycho (changes requested):
I think we would want to add a admission policy webhook for validating the URL, and maybe doing some sanitization on it

JoseSzycho:
@mattdjenkinson there are some conflicts

PR #749, opened 2026-08-05
fix: scope suspension events via impersonated parent context (#748)

### LINKED ISSUE #748: Suspension events are invisible to consumer activity feeds (blocks cloud-portal#1356 AC #2)
> [!NOTE]
> **Updated 2026-08-06.** The fix direction changed after @scotwells' review below: scope must come from the request's parent context via impersonation, not from a second derivation path in the events proxy. The original description is struck through at the bottom.

## Problem

Controller-emitted `Suspended`/`Reinstated` events never reach consumer activity feeds.

Scope annotations (`platform.miloapis.com/scope.type` / `scope.name`) are derived only from the requesting user's `iam.miloapis.com/parent-type` / `parent-name` extras. The propagator controller emits through the manager's shared EventRecorder, which authenticates as the controller and carries no parent context — so these events are stored untagged, and project-scoped feed queries (which filter strictly on scope, with no hierarchy resolution) can never return them.

## Approach

Set the parent on the request via impersonation, so the single existing scope-injection path works unchanged:

- Emit through a client that impersonates the controller's service account plus `parent-type: Project` / `parent-name: <project>`.
- Do **not** add a second way to derive scope annotations.

Confirmed viable: `WithImpersonation` is enabled in Milo's handler chain (`cmd/milo/apiserver/config.go:525`) and runs *before* the Project/Organization context decorators, which pass through untouched on requests that aren't project-scoped.

## Also in scope

1. Carry the suspension reason **category** in the event and policy summary. Never the ProjectSuspension resource name, `requestedBy`, or description. The same redaction applies to `Project.status.conditions["Suspended"].message`, which project members read directly.
2. Make the `Project.status.suspensions[]` phase contract explicit — document active-only, filter by phase, or add phase to `ProjectSuspensionInfo`.

## Known prerequisite

The controller-manager authenticates as `admin` / `system:masters` in test-infra, so impersonation RBAC is not exercised there. Moving it onto its service account identity is a separate migration — it changes authentication for every controller in that binary at once, and would expose any existing gaps in the generated ClusterRole.

---

<details>
<summary><b>Original description (superseded)</b></summary>

~~Part of the **Project Suspension & Lifecycle Controls** enhancement (datum-cloud/enhancements#800).~~
~~Blocks the consumer suspension timeline in datum-cloud/cloud-portal#1356 (AC #2).~~

## ~~1. Controller-emitted Suspended/Reinstated events carry no tenant scope~~

~~The propagator controller emits the consumer-facing suspension events~~
~~(`internal/controllers/resourcemanager/metrics.go:99,106` —~~
~~`eventRecorder.Eventf(project, "Normal", "Suspended", ...)` / `"Reinstated"`).~~

~~Activity scope annotations (`platform.miloapis.com/scope.type` / `scope.name`) are injected~~
~~only from the *requesting user's* `iam.miloapis.com/parent-type` / `parent-name` extras~~
~~(`internal/apiserver/events/scope.go` — `injectScopeAnnotations`, single injection site at~~
~~`eventsv1_rest.go:297`). A controller has no parent context, so these events are stored~~
~~**untagged**. Project-scoped activity queries filter `scope_type = 'project' AND scope_name = ?`,~~
~~and scopes are not hierarchically inclusive (activity `docs/architecture/multi-tenancy.md`) —~~
~~so the events the `project-policy.yaml` eventRules were written for~~
~~("Project {{name}} was suspended") can never appear in a consumer's project feed.~~

~~**Ask:** tag controller-emitted Project lifecycle events with the involved project's scope~~
~~(e.g. derive from `involvedObject` for this event path, or emit through a context carrying~~
~~the project parent).~~

~~Additionally, the eventRule summaries carry no reason category, which the consumer timeline~~
~~needs ("Suspended — reason: abuse"). The reason *category* is consumer-safe by design~~
~~(enhancement README "Show the consumer the reason, not the internals"); please include it~~
~~in the event/summary. Do **not** include the ProjectSuspension resource name, requestedBy,~~
~~or description.~~

## ~~2. Make the Project.status.suspensions[] phase contract explicit~~

~~`projectsuspension_propagator_controller.go` appends **every** ProjectSuspension targeting~~
~~the project to `status.suspensions[]` regardless of `status.phase`, and~~
~~`ProjectSuspensionInfo` carries no phase field. Today this is effectively active-only because~~
~~lift is implemented as delete (`PhaseLifted` is never written by non-test code, and~~
~~staff-portal lifts by deleting the resource). If lift ever becomes a phase transition instead~~
~~of a delete, every consumer of `status.suspensions[]` silently starts rendering lifted~~
~~suspensions as active.~~

~~**Ask:** either document active-only as the contract of `status.suspensions[]`, or filter by~~
~~phase in the propagator (matching the existing `activeSuspensionNames` logic), or add the~~
~~phase to `ProjectSuspensionInfo`.~~

</details>

yahyafakhroji:
Fixes #748.

## What

**1. Suspension events now carry tenant scope, set via impersonation.**

Scope annotations (`platform.miloapis.com/scope.*`) are derived from the requesting user's `iam.miloapis.com/parent-type` / `parent-name` extras. The propagator emitted through the manager's shared EventRecorder, which authenticates as the controller and carries no parent context — so its events were stored untagged and invisible to project-scoped feed queries.

`ProjectEventEmitter` emits through a client that impersonates the controller's service account plus `parent-type: Project` / `parent-name: <project>`, so `injectScopeAnnotations` takes its existing full-parent-context branch unchanged. Per @scotwells' review, no second scope-derivation path.

Verified at the code level: `WithImpersonation` is enabled in Milo's handler chain (`cmd/milo/apiserver/config.go:525`) and runs *before* the Project/Org context decorators, which pass through untouched when the request isn't project-scoped (`pkg/server/filters/projects.go:86-91`).

Events are created as `events.k8s.io/v1` so they genuinely carry the `Regarding`/`Note` fields the ActivityPolicy reads, with no conversion in between.

**2. Internal admin detail redacted.** The `Suspended` event message and `Project.status.conditions["Suspended"].message` both embedded ProjectSuspension resource names. Both now carry only the reason category (e.g. `Abuse`). Project status is tenant-readable, so it needed the same treatment as the event.

**3. ActivityPolicy CEL fields corrected.** `eventRules` read `event.involvedObject.name` / `event.message`; the `events.k8s.io/v1` contract exposes `event.regarding.name` / `event.note`. The summary would have silently rendered empty.

**4. `status.suspensions[]` filtered to active phase**, with the active-only contract documented on the API types. No-op today since lift is implemented as delete.

**5. RBAC** regenerated from new kubebuilder markers: `impersonate` on `serviceaccounts` and on the two `userextras` parent keys, plus `create` on `events.k8s.io/events` — the generated ClusterRole had no events-create rule in any API group.

## Notes for reviewers

**The involved-object scope fallback is still in the diff.** I kept it rather than removing it here: impersonation is verified at the code level but has not yet run at runtime (see below). The two coexist safely — impersonation supplies full parent context, so the fallback branch is never reached for these events. Happy to remove it if you'd rather the PR show only the impersonation path.

**The impersonation RBAC is not exercised by CI.** The controller-manager authenticates as `admin` / `system:masters` in test-infra, so a green run there will not prove the new rules are correct. Moving it onto its service account identity changes authentication for every controller in that binary at once and would surface any other gaps in the generated ClusterRole, so it belongs in its own change.

## Test plan

- `go build ./...`, `gofmt -l .` — clean
- `go vet ./...` — one pre-existing, unrelated finding (`organization_types.go:127`), confirmed present on `main`
- `go test -timeout 5m ./...` — 772 passed, 72 packages

### DIFF
--- a/config/controller-manager/overlays/core-control-plane/rbac/role.yaml
+++ b/config/controller-manager/overlays/core-control-plane/rbac/role.yaml
@@ -17,6 +17,12 @@ rules:
   - update
   - watch
 - apiGroups:
+  - ""
+  resources:
+  - serviceaccounts
+  verbs:
+  - impersonate
+- apiGroups:
   - '*'
   resources:
   - '*'
@@ -39,6 +45,13 @@ rules:
   verbs:
   - patch
   - update
+- apiGroups:
+  - authentication.k8s.io
+  resources:
+  - userextras/iam.miloapis.com/parent-name
+  - userextras/iam.miloapis.com/parent-type
+  verbs:
+  - impersonate
 - apiGroups:
   - authorization.k8s.io
   resources:
@@ -65,6 +78,12 @@ rules:
   - patch
   - update
   - watch
+- apiGroups:
+  - events.k8s.io
+  resources:
+  - events
+  verbs:
+  - create
 - apiGroups:
   - gateway.networking.k8s.io
   resources:
--- a/config/crd/bases/resourcemanager/resourcemanager.miloapis.com_projects.yaml
+++ b/config/crd/bases/resourcemanager/resourcemanager.miloapis.com_projects.yaml
@@ -136,11 +136,19 @@ spec:
                   type: object
                 type: array
               suspensions:
-                description: Suspensions lists the active/all suspensions currently
-                  affecting the project.
+                description: |-
+                  Suspensions lists the active suspensions currently affecting the
+                  project. See ProjectSuspensionInfo for the active-only contract.
                 items:
-                  description: ProjectSuspensionInfo contains the details of a suspension
-                    affecting the project.
+                  description: |-
+                    ProjectSuspensionInfo contains the details of a suspension affecting the
+                    project.
+
+                    Contract: only ProjectSuspensions in an active phase (status.phase ==
+                    "Active", or unset/not-yet-reconciled) are represented here. A
+                    ProjectSuspension that has moved past Active (e.g. "Lifted") is excluded.
+                    Consumers must not assume every entry in status.suspensions still applies
+                    enforcement/visibility beyond that active-only guarantee.
                   properties:
                     reason:
                       description: Reason is the category of suspension.
--- a/config/services/activity/policies/resourcemanager/project-policy.yaml
+++ b/config/services/activity/policies/resourcemanager/project-policy.yaml
@@ -10,6 +10,15 @@
 # - Use the project name as display text
 # - Action-oriented language ("created project", "deleted project", "was suspended")
 # - Exclude system actors to reduce reconciliation noise
+#
+# eventRules for Suspended/Reinstated use event.regarding.name and
+# event.note (the events.k8s.io/v1 CEL contract exposes the Kubernetes
+# Event's involved object as `regarding`, not `involvedObject`, and its
+# message as `note`, not `message` — see datum-cloud/activity
+# internal/cel/environment.go). The controller crafts that note to be
+# consumer-safe: it names the suspension reason CATEGORY (e.g. "Abuse") but
+# never the ProjectSuspension resource name, requestedBy, or description,
+# which are internal admin detail.
 apiVersion: activity.miloapis.com/v1alpha1
 kind: ActivityPolicy
 metadata:
@@ -38,8 +47,8 @@ spec:
   eventRules:
     - name: suspended
       match: "event.reason == 'Suspended'"
-      summary: "Project {{ event.involvedObject.name }} was suspended"
+      summary: "Project {{ event.regarding.name }} was suspended (category: {{ event.note }})"
 
     - name: reinstated
       match: "event.reason == 'Reinstated'"
-      summary: "Project {{ event.involvedObject.name }} was reinstated"
+      summary: "Project {{ event.regarding.name }} was reinstated"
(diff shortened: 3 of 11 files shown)


internal/apiserver/events/scope.go
JoseSzycho, line 1: @scotwells could you review this file? All other files, except this and `internal/apiserver/events/eventsv1_rest.go` looks correct.

internal/apiserver/events/eventsv1_rest.go
JoseSzycho, line 1: @scotwells could you take a look at this file? everything else, expect this file and `internal/apiserver/events/scope.go` looks correct

scotwells:
@yahyafakhroji we should not fallback to annotations. They're untrusted. Upstream controllers must always use impersonation to authenticate to a project.

yahyafakhroji:
@scotwells Agreed — fallback removed entirely. Scope now comes only from the request's parent context; nothing on the event body can influence it, and the old fallback tests are inverted to assert exactly that. The suspension controller impersonates `parent-type: Project` / `parent-name: <project>` to emit. Net −189 lines.

Separately: the quota controllers emit via the shared EventRecorder with no parent context, so they're unscoped too — happy to open a follow-up.

PR #756, opened 2026-08-10
feat(iam,admission): email verification gate + idempotent User admission

yahyafakhroji:
## What's the current problem?

Nothing on the API side stops an account that never confirmed its email from creating things. The portal can redirect someone to "check your inbox", but that's a UI redirect — anyone holding a token can call the API directly.

Separately, duplicate `User` creates return the wrong error. The admission webhook creates three side-effect resources with no `AlreadyExists` tolerance, and Kubernetes runs validating admission *before* the etcd write — so a retried or concurrent signup fails admission on the `User` itself. zitadel-provider's `EnsureUser` tests `IsAlreadyExists`, a predicate an admission denial never satisfies, so benign retries surface as 500s.

## How are we solving it?

- **An admission plugin** reading `iam.miloapis.com/emailVerified` off the authenticated identity, stamped by [zitadel-provider#127](https://github.com/milo-os/zitadel-provider/pull/127) beside the existing `registrationApproval` key.
- **Idempotent `User` admission**, so a duplicate create returns `AlreadyExists`. The `UserPreference` site re-reads on conflict rather than merely tolerating it: `Create` leaves UID empty, and the third `PolicyBinding` would stamp that empty UID into its `ResourceRef`.

**Nothing enforces on merge.** `EmailVerifiedGate` defaults off, which means *observe*, not *disabled* — see below.

## Why are we solving it?

Part of the passkey program — [enhancements#738](https://github.com/datum-cloud/enhancements/issues/738), tracked for this phase in [auth-ui#111](https://github.com/datum-cloud/auth-ui/issues/111). An unverified account shouldn't be able to act on the platform.

## Why a plugin rather than a policy

A `ValidatingAdmissionPolicy` expressing the same rule was carried in parallel ([datum#268](https://github.com/datum-cloud/datum/pull/268)) while this was open. Two things it cannot do:

- **Emit `EmailNotVerified` in `details[].causes`.** A policy's `reason` is limited to the built-in enum, so cloud-portal would match message text — where it currently matches `details[].code` for suspension.
- **Reach project control planes.** A policy is deployed to the control planes the GitOps repo targets; `config/crd/bootstrap.go` seeds CRDs, not policies.

## Three things worth knowing

**This is 60 lines, not the 559 it started as.** The first version read `User.status.emailVerified`, which cost a per-write API call, an informer cache of ServiceAccount UIDs to tell machines from humans, a readiness check, and a CI timeout when the reflector listed a CRD the bootstrap hook hadn't created. Moving the signal onto the identity removed all of it: a machine token carries no `email` claim, which is the same discriminator `EffectiveUsername()` already uses.

**`EmailVerifiedGate` selects deny-vs-observe, not on-vs-off.** The plugin always evaluates and always counts; disabled, it admits anyway. Read `milo_email_verification_denials_total{enforced="false"}` before enabling — that's the population that starts being denied, and it's what makes the rollout possible without a backfill Job. The label mirrors a policy's `enforcement_action` so the numbers stay comparable if both ever run.

**Absence admits, as a contract.** Machine identities carry no key and pass. zitadel-provider guarantees the converse and tests it: a human *always* carries the key, `"true"` or `"false"`, never omitted. If that breaks, this fails open for humans — the two sides change together.

## Verification

`go build`, `go vet`, `gofmt` clean; 30 test packages pass. Plugin cases cover: verified admits, unverified denies on create and update, absent key admits, empty value slice admits, unexpected value denies, access review exempt, `system:masters` exempt, mid-deletion update exempt, create-with-nil-`oldObject` still denies, the cause type is present, `Delete`/`Connect` unregistered, observe mode counts without denying, and the gate defaults off.


JoseSzycho:
same as here https://github.com/milo-os/zitadel-provider/pull/127#issuecomment-5240569064

yahyafakhroji:
> same as here https://github.com/milo-os/zitadel-provider/pull/127#issuecomment-5240569064

@JoseSzycho sorry for that,, i just updated it

JoseSzycho (changes requested):
@yahyafakhroji Did you consider using a [ValidatingAdmissionPolicy](https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/)`instead of writing a plugin and maintaining a cache?

Here's an example of one we're already using:
https://github.com/datum-cloud/datum/blob/main/config/services/iam.miloapis.com/validation/approved-user-policy.yaml

This would make the policy much easier to maintain.

Here's the code in the `zitadel-provider` repo that populates the information used by this `ValidatingAdmissionPolicy`:

https://github.com/milo-os/zitadel-provider/blob/171a2e7c017fb5dc04d042604b14059117af42d4/internal/webhook/response.go#L21-L41

---

Apart from that, I noticed that this PR also adds an Email TTL for automatically deleting emails. Please move this to a separate PR, since it's unrelated to the rest of the changes in this PR.

This is something we need, and we've already created an issue for it:
https://github.com/milo-os/resend-provider/issues/33

Could you also move the controller to the `resend-provider` repo?

In the medium term, this controller logic should live in its own repository, where we'll keep notification logic that doesn't implement a third-party service. In the meantime, though, it would be fine to have the controller in `resend-provider`.

yahyafakhroji:
> [!NOTE]
> **Mostly resolved — kept for history.** Struck-through parts are done or superseded; one item is still live at the bottom.

~~Thanks @JoseSzycho — agreed on both.~~

~~**Email TTL** — splitting it out. The API field has to stay in milo (the CRD lives here); the controller goes to `resend-provider` against milo-os/resend-provider#33. One check: #33 mentions cleanup at a size threshold, but I built TTL-after-delivery plus a max-age ceiling, with no size trigger. Enough for now?~~

→ **Superseded, and I had this wrong.** The field left milo as well. Retention isn't among [enhancements#738](https://github.com/datum-cloud/enhancements/issues/738)'s goals and [#33](https://github.com/milo-os/resend-provider/issues/33) is storage-motivated, so both the controller *and* `spec.ttlSecondsAfterDelivery` were descoped rather than split. The size-threshold question moves to #33's own track.

~~**VAP** — you're right. Most of the plugin's weight is a ServiceAccount cache, and it only exists because the signal lives in a CR that's created asynchronously — there's a window where a human has a valid token but no `User` yet. Putting `emailVerified` in `userInfo.extra` like `registrationApproval` closes that window, and the cache goes with it. Cheaper on the provider side too.~~

→ **Done.** [datum#268](https://github.com/datum-cloud/datum/pull/268) (policy) + [zp#127](https://github.com/milo-os/zitadel-provider/pull/127) (producer).

~~**One question before I rewrite it:** the plugin runs in every milo apiserver, including project control planes. A VAP is deployed through `datum/config/services/...` and milo's bootstrap seeds CRDs only — so project control planes look uncovered. `deny-unapproved-user` has the same gap today. Accepted limitation, or is there a distribution path I'm missing?~~

→ **Answered** — plugin dropped, the policy is the only enforcement. Worth noting milo owns no `ProjectControlPlane` controller (only the CRD type), so how those apiservers get configured was never milo's to decide anyway.

---

**Still live:** the plugin denied with a custom `EmailNotVerified` cause in `details[].causes`, which a policy can't emit — `reason` is limited to the built-in enum. cloud-portal matches suspension 403s on `details[].code`, so the verification equivalent will have to match message text. Tracking on [cp#1430](https://github.com/datum-cloud/cloud-portal/pull/1430), not here.

internal/apiserver/admission/plugin/emailverification/admission.go
```diff
@@ -0,0 +1,183 @@
+// Package emailverification implements the EmailVerificationEnforcement
```
JoseSzycho, line 1: we don't need the admission plugin as you already created the AdmissionValidatingPolicy
https://github.com/datum-cloud/datum/pull/268/changes#diff-4021d1b1ed95be0edd8159d93e70bd25757e2e38d4a20685eaaf4cb15cdd5a55

yahyafakhroji:
@JoseSzycho  Done — plugin removed in 4fb042ea, so enforcement lives only in datum#268.

JoseSzycho (approved):

PR #758, opened 2026-08-11
feat: Escalate suspended projects to deletion after a configurable retention window

### LINKED ISSUE #721: Reclaim abandoned suspended projects (retention → escalation)
A project left suspended past a time-boxed window escalates to (reversible, time-boxed) deletion — bounding resource hoarding while keeping suspension itself non-destructive.

> Fast-follow: not launch-blocking.

## Acceptance criteria
- [ ] Configurable retention window; escalation to deletion after it elapses.
- [ ] The window and its notifications are explicit and auditable.
- [ ] Escalation is the only suspension → deletion path; clearly communicated to project owners.

**Dependencies:** platform primitive.

---
Part of the **Project Suspension & Lifecycle Controls** enhancement (datum-cloud/enhancements#800). Rendered proposal: https://github.com/datum-cloud/enhancements/blob/enhancement/project-suspension/enhancements/platform/project-suspension/README.md

JoseSzycho:
## Description

Suspending a project today is permanent unless an operator manually reinstates it. That's a problem: a suspended project still holds onto its resources indefinitely, with nothing forcing a resolution. This PR adds an automatic escalation path — if a project stays suspended too long, it gets deleted, following the same cleanup flow that already runs for normal project deletion (finalizer + resource purge). Nothing new is invented for the actual deletion mechanics; this just decides *when* to trigger it.

### How it works

A new controller, `ProjectSuspensionEscalationController`, watches every `Project`. The moment a project's `Suspended` condition flips to `True`, it computes a deletion deadline (today + the configured retention window) and writes it to `status.suspensionEscalation.deletionAt`, along with a `PendingDeletion` condition so it's visible on the resource itself. From then on, every reconcile checks how much time is left. Once the deadline passes, the controller deletes the `Project` — that's it, the existing deletion controller takes over from there exactly as it would for a manual delete. If the project gets reinstated before the deadline, the escalation state is cleared and nothing happens.

### How the emails work

The point of escalation isn't to surprise anyone, so the controller also keeps the project's organization in the loop by e-mail. It sends one the moment the project is suspended, then again as the deadline gets closer, at whatever "days remaining" checkpoints are configured. Each e-mail is built the same way the rest of the platform sends transactional e-mail — it creates an `Email` resource pointing at an `EmailTemplate`, and the existing email pipeline takes it from there. The recipient is whatever contact e-mail is set on the organization; if there isn't one, the controller skips sending (and logs it) rather than failing. E-mails are written to a fixed namespace (`milo-system` by default) instead of the organization's own namespace, so they stick around as an audit trail even if the organization or project is later cleaned up. Every checkpoint is recorded on the project's status so the same warning is never sent twice, even across reconciles.

### What's configurable

Everything about the timing and templates is a controller-manager flag, so operators can tune it without a code change:

- `--project-suspension-retention-window-days` — how many days a project can stay suspended before it's deleted.
- `--project-suspension-notification-days` — a comma-separated list of "days remaining" checkpoints to send a reminder at, on top of the initial notice sent at suspension time.
- `--project-suspension-deletion-warning-email-template` — which `EmailTemplate` to use.
- `--project-suspension-deletion-warning-email-namespace` — where the sent `Email` resources are stored.

All four have sensible defaults and are validated at startup — a zero or negative retention window, or a zero or negative notification day, fails fast instead of silently misbehaving.

---

Related to:
- https://github.com/milo-os/milo/issues/721

Main enhacement:


JoseSzycho:
Set to draft. The email template must be updated.

https://datum-inc.slack.com/archives/C084W6XRVS7/p1786404439122749

scotwells:
Should we always be sending these warning emails? Thinking of scenarios where a project is suspended for fraud and they may lose access to their account entirely. Would be good to confirm what communications we want sent out for the different types of project suspensions that may happen. Should be able to work with the GTM team on this.

JoseSzycho:
@scotwells great idea! Thanks for the review.

Will work on this with the GTM team on this.

JoseSzycho:
@scotwells according to some conversation with the GTM team, I made the required changes in the controller.

We'll avoid sending warning emails caused by Fraud/Abuse reasons (this can be modified by flags).

Additionally, I created this issue, so staff portal operator can have a better understanding of ongoing project suspensions:
- https://github.com/datum-cloud/staff-portal/issues/623

JoseSzycho:
@gaghan430 as per your request, I added a way for filtering project-suspension related emails.

- https://github.com/milo-os/milo/pull/758/commits/c32de3100b26d3874d8c937bb03696aa1769a507

scotwells (approved):

JoseSzycho:
@mattdjenkinson thanks for that review.

For the issue 1, this is correct, but is a border scenario for already suspended projects, and this is a new feature, and not real project has been yet suspended, I think we can ignore it .

For issue 2, yes, there is a gap there. If somehow our controller manager stop working for some days, they remaining days mentioned in the emails will have an offset. I'm fixing that in a new PR.

Thanks for it!

PR #759, opened 2026-08-11
test(resourcemanager): disable global configmap quota policy during project deletion blocked resources e2e test

JoseSzycho:
## Summary

Fixes e2e test failure in `chainsaw/project-deletion-blocked-resources`.

### Cause
When the test creates `ConfigMap` resources (`held-resource`) inside project control planes, Milo's global quota policy (`project-configmap-claim-policy`) intercepts the request and generates a `ResourceClaim`. Because the test organization has no `ResourceGrant` allocated for `core.miloapis.com/configmaps`, admission denies resource creation with `403 Forbidden` (`You've reached your quota for this resource type`).

### Solution
- Temporarily patch `project-configmap-claim-policy` to `disabled: true` during `clear-previous-run` in `chainsaw-test.yaml`.
- Re-enable the policy (`disabled: false`) during test cleanup.

## Testing
- Verified locally with `task test:end-to-end -- resource-management/project-deletion-blocked-resources` (Passed: 1, Failed: 0).

### DIFF
--- a/test/resource-management/project-deletion-blocked-resources/chainsaw-test.yaml
+++ b/test/resource-management/project-deletion-blocked-resources/chainsaw-test.yaml
@@ -66,6 +66,9 @@ spec:
               kubectl --kubeconfig kubeconfig-project-namespace -n blocked-ns patch configmap held-resource \
                 --type=merge -p '{"metadata":{"finalizers":null}}' >/dev/null 2>&1
 
+              kubectl --kubeconfig kubeconfig-main patch claimcreationpolicy project-configmap-claim-policy \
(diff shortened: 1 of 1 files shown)


scotwells (approved):
Seems odd we have to disable it like this? Why are the default quota policies enabled by default anyways? maybe something we need to re-organize?

JoseSzycho:
> Seems odd we have to disable it like this? Why are the default quota policies enabled by default anyways? maybe something we need to re-organize?

yeah, it is odd! I tackled this quickly as I saw many other PRs failing because of this problem.

I don't know if it is related to some residual resource of the quota e2e tests? I'm not sure. 

I'm creating an issue so we can keep track of this.

PR #777, opened 2026-09-06
chore: Restore e2e and test-doc tasks on a current Go toolchain

ecv:
## Summary

The pinned chainsaw release does not build on Go 1.26 or newer, so the e2e and test-doc tasks fail on a current local toolchain.

This change moves the pin to v0.2.15, which carries the upstream fix. The new release declares Go 1.26, so the module now names that toolchain and the builder image moves to the same version, which lifts every CI job and local build from Go 1.25 to 1.26.

Two generated test READMEs are refreshed too, one stale since August.

## Test plan

- [ ] The e2e suite installs the new release and passes
- [ ] Regenerating test docs on this branch produces no diff

Related to #779

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@
 # and cross-compiles to $TARGETOS/$TARGETARCH. This makes multi-arch builds
 # (linux/amd64, linux/arm64) fast under buildx without requiring QEMU emulation
 # for the build itself.
-FROM --platform=$BUILDPLATFORM golang:1.25 AS builder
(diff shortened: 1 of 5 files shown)


kevwilliams (approved):
Chainsaw pin bump (v0.2.13 → v0.2.15) correctly resolves the Go 1.26 build failure, and the go.mod toolchain directive plus Dockerfile builder bump are consistent with it. The builder image bump does touch the production build image, not just test/e2e, but that's unavoidable with a single go.mod and is disclosed in the description. Regenerated READMEs match the new chainsaw step counts. CI is fully green.

PR #778, opened 2026-09-06
fix: Reject spec-less PlatformAccess and UserPreference

### LINKED ISSUE #775: PlatformAccess and UserPreference accept a spec-less object that never defaults
## What happens

`PlatformAccess` and `UserPreference` in `iam.miloapis.com/v1alpha1` have an optional top-level `spec` with required sub-fields and nested defaults. An object applied with no `spec` key is accepted, and none of the nested defaults apply, because CRD defaulting fills a nested default only when the parent object is present.

This is the same defect class as #773, which #774 fixes for `ServiceAccount` only. Found during review of #774.

## Fix

For each type, pick one:

- Add `+kubebuilder:default={}` to the `Spec` field and regenerate, so an absent spec resolves to an empty object and the nested defaults apply.
- Mark `spec` required in the schema, so an absent spec is rejected at admission.

Add a schema test beside the one #774 adds for `ServiceAccount` so the choice cannot regress.

Related to #773.

ecv:
## Summary

PlatformAccess and UserPreference accepted objects with no spec, and none of their nested defaults applied.

This change marks spec required on both kinds. Both require a user reference with no default, so an empty spec default, the route #774 took, would still produce an invalid object.

## Breaking changes

Existing stored objects of either kind with no spec become unwritable once the schema lands, status writes included.

A live listing found 443 PlatformAccess objects, all with spec. UserPreference could not be listed with the credentials at hand. The only creator in the tree is the User webhook, which always sets the user reference and theme, and the manager role cannot create, so no stored UserPreference is expected to lack spec. The author accepts that reasoning in place of a count.

## Test plan

- [ ] A UserPreference applied without spec is rejected at admission
- [ ] A stored PlatformAccess cannot have spec removed by update

Fixes #775
Related to #774


kevwilliams (approved):
Sound fix. Marking spec Required via kubebuilder marker (and regenerated CRD schema) rejects spec-less objects at admission through the apiserver's native OpenAPI validation, so there's no custom webhook logic or nil-pointer risk. Blast radius is well-analyzed: a live listing confirmed all 443 PlatformAccess objects have spec, and the UserPreference creation path (User webhook) always sets required fields. Test coverage checks the generated CRD schema enforces required+no-default correctly for both kinds, which is the right level given enforcement is delegated to the apiserver.

PR #782, opened 2026-09-08
feat(iam): EmailVerification status field, PasskeyRegistrationLink type, and recovery-link roles

yahyafakhroji:
## What

First of two milo PRs for Phase C of the passkey program (account recovery). API and authorization objects only; no controller behaviour changes.

- **`EmailVerification` status field on `User`** — `UserStatus.EmailVerification` of type `EmailVerificationState`, `Verified` | `Unverified`, declared alongside `State` and `PlatformAccess`. Empty means zitadel-provider has not synced the user yet, so readers must treat anything but `Verified` as unproven. Exposed as the selectable field `.status.emailVerification`. Written ONLY by zitadel-provider — on provisioning, on the `user.human.email.verified/changed` events, and reconciled by its sweeper; milo controllers only read it.
- **`PasskeyRegistrationLink` identity API type** — `Spec{UserRef, RequestedBy, Reason}`, `Status{UserUID, EmailName, ExpiresAt}`. Served by zitadel-provider's identity aggregated apiserver as a create-only virtual resource; the durable record is the notification `Email` it produces.
- **`ProtectedResource` + roles** `identity-passkey-registration-links-{viewer,editor,admin}`, mirroring `protected-resources/iam/userdeactivation.yaml` and the service-account-keys roles. Dormant until an infra `PolicyBinding` binds them to people; the activation switch is zitadel-provider's `--recovery-links-enabled` flag.

## Why

Gap 2 of milo-os/zitadel-provider#134: an account that never verified its email still receives the welcome mail, because nothing on `User` says whether the address is proven. The field gives controllers that signal (PR 2 gates the welcome and rejection mails on it). The link type and roles are the admin backstop for datum-cloud/auth-ui#112 (support sends a passkey recovery link from staff-portal).

## Notes for reviewers

- Amended after review on zitadel-provider#137: typed status field instead of a condition, mirroring State/PlatformAccess.
- `task generate:test-docs` fails on `main` before this branch (pre-existing, unrelated).
- `task generate:openapi:identity` rewrites dotted `meta/v1` definition names; the generated file was restored to the committed dotted form by hand. A separate branch addresses the generator.
- `go vet` reports a pre-existing finding in a file this PR does not touch.

## After merge

Please tag. zitadel-provider's Phase C branch pins `go.miloapis.com/milo` to that tag; PR 2 (waitlist controller waits for `EmailVerification: Verified`) follows once zitadel-provider has completed one sweep in staging.


pkg/apis/identity/v1alpha1/passkeyregistrationlink_types.go
```diff
@@ -0,0 +1,62 @@
...
+// PasskeyRegistrationLinkSpec is what support asks for.
+type PasskeyRegistrationLinkSpec struct {
+	// UserRef names the iam.miloapis.com User (metadata.name is the provider user ID)
+	// who receives the link. The link only ever goes to that user's verified address.
+	UserRef PasskeyRegistrationLinkUserReference `json:"userRef"`
+	// RequestedBy is the metadata.name of the staff User asking. Recorded for audit;
+	// the server rejects a value that does not match the authenticated caller.
+	RequestedBy string `json:"requestedBy"`
+	// Reason is why the link is being sent (ticket reference, customer request). Required.
+	Reason string `json:"reason"`
```
JoseSzycho, line 40: if required, you should add the corresponding kubebuilder marker

JoseSzycho (changes requested):

pkg/apis/identity/v1alpha1/passkeyregistrationlink_types.go
yahyafakhroji: Good catch, thanks. Added `+kubebuilder:validation:Required` on the spec fields (userRef, its name, requestedBy, reason) and explicit `Optional` on the status fields, matching the other identity types. The generated OpenAPI already listed them as required (it follows the json tags), so no generated file changed. One extra commit on top so nothing else moves.

JoseSzycho (changes requested):
please run `task generate`, as the new user.status needs to update the API doc

JoseSzycho (approved):

PR #785, opened 2026-09-10
fix: Declare two correct rules empty by design

ecv:
## Summary

Production reports the controller-manager crash-loop warning and the membership readiness alert as rules that have selected no series for over a week, which reads exactly like a rule whose metric was renamed or never shipped.

Both are correct as written. The crash-loop rule selects series only while the pod is in CrashLoopBackOff, and the readiness rule counts only memberships that are not Ready, reading 547 membership series on production and emitting nothing.

Both now carry a label declaring the emptiness intended, and the membership runbook records why the rule looks dead when the fleet is healthy.

## Test plan

- [ ] The no-data report on production stops naming these two rules after the next release reaches it
- [ ] A crash-looping controller and an unready membership each still raise their alert

Related to https://github.com/milo-os/milo/issues/783
Related to https://github.com/datum-cloud/infra/issues/5086

https://claude.ai/code/session_014zQEpY3TeCaSfH7nEY5Mzg

### DIFF
--- a/config/telemetry/alerts/controller-manager/controller-manager.yaml
+++ b/config/telemetry/alerts/controller-manager/controller-manager.yaml
@@ -35,6 +35,7 @@ spec:
             }[5m]) >= 1
           for: 5m
           labels:
+            expected_no_match: "true"
             severity: critical
             service: milo-controller-manager
           annotations:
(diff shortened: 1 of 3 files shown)


kevwilliams (approved):
Confirmed both rules are correct as written: the crash-loop alert only selects series while a pod is in CrashLoopBackOff, and the membership readiness rule only counts non-ready memberships out of 545-547 live series, so a healthy fleet legitimately produces zero series rather than a bug. The `expected_no_match: "true"` label is applied to exactly the right alert in each file (not the unrelated Restarting alert), matches the documented compliance-exporter convention, and the runbook addition records the reasoning for future readers. Scope is tight and matches the linked issues.

PR #789, opened 2026-09-16
feat(iam): Show the member email in group activity

yahyafakhroji:
**Problem**
The activity log shows a user ID when someone is added to or removed from a group, because the GroupMembership only carries the User name and the audit event never contains the email. The activity policy had nothing better to print.

**Solution**
A mutating webhook on GroupMembership looks up the referenced User at admission and stamps its email into the `iam.miloapis.com/user-email` annotation. It only decorates: a missing User leaves the object untouched and the webhook runs with `failurePolicy: Ignore`, so an outage can never block adding a member. The activity policy prefers the annotation and falls back to the User name, so memberships created by other clients still render.

OrganizationMembership has the same shape and can take the same annotation as a follow-up.

Refs datum-cloud/cloud-portal#1507


JoseSzycho (changes requested):
Good job. Remember that the defaulter runs only once.

If the user change their email address (which is already supported), the information in the annotation will get stale.

I would suggest you to update the groupmembersip with a new status field that for storing the user email address.

After that, update the corresponding controller, so every time the user updates their email address, the group membership receive the corresponding email address update.

Please, do not forget to also update the e2e test, and verify they are green before assigning the PR to review.

yahyafakhroji:
Thanks @JoseSzycho , agreed on the staleness. Proposal before I rework it:

- Add `status.userEmail` (plus `observedGeneration`) to GroupMembership, owned by a new GroupMembership controller that mirrors the OrganizationMembership one: index on `spec.userRef.name`, watch Users, re-reconcile on email change.
- Keep the admission stamp only because the create audit event has no status yet (status is dropped on create and a webhook can't set it), so the activity line for "added X" would still show the ID without it. The annotation then means "email when the member was added", and status means "current email"; the controller never touches the annotation.
- Activity policy: create reads the annotation, delete reads `status.userEmail` with the annotation as fallback.
- Move the e2e into its own chainsaw suite without the `authorization-provider` label so it actually runs in CI, and make sure it's green before re-requesting.

OrganizationMembership can take the same status field afterwards since its controller already watches Users. Does that work for you?

JoseSzycho:
@yahyafakhroji you requested my review, but I see no changes

yahyafakhroji:
@JoseSzycho I haven't done any updates yet, I just open discussion for my propose https://github.com/milo-os/milo/pull/789#issuecomment-5707341703

JoseSzycho:
@yahyafakhroji I understand that you need the resource to be created with the annotation, and that the status get dropped. AS you need to be able to read the email address at creation time, instead of adding a new status field, it would be fine that your new controller always keep the annotation updated when the email address changes.

So we would only work around that new annotation field.

You should move the `UserEmailAnnotation` key to the type definition of the resource. So other repos can import it if they need it.

yahyafakhroji:
Thanks @JoseSzycho, the annotation-only approach is cleaner. Dropped the status field.

What changed:

- `UserEmailAnnotation` moved to `pkg/apis/iam/v1alpha1/groupmembership_types.go` so other repos can import it.
- New `GroupMembershipController` indexes memberships on `spec.userRef.name`, watches Users, and patches just the annotation when an email changes. A missing user or an empty email leaves the membership untouched, same as the webhook.
- The old e2e step was in `test/group`, which CI skips because the suite is labelled `requires: authorization-provider`, so it never actually ran. Added `test/iam/groupmembership-user-email/` without that label, covering both the create and the email change.

scotwells (changes requested):
This would be a better use-case for control plane events instead of an audit log being the source of the annotation. You would just have the system emit an event saying the user was added to the group when it processes its group membership.

yahyafakhroji:
@scotwells that makes sense, an event means we don't store the email on the resource at all.

Two things I'd need to settle first:

- **Actor.** Audit rules get the human from `audit.user.username`. Event rules resolve the actor from `reportingController`, so the line would read as the controller instead of "Yahya added ...". The membership doesn't record who created it. How should the controller know the actor?
- **Deletion.** "Removed X from group Y" is a delete audit entry today. Once the object is gone there's nothing left to reconcile, so would the removal event need a finalizer, or is there an existing pattern for that?

scotwells:
Maybe we look at updating the activity service to support enriched context by querying other API resources so it can resolve the user referenced in the group membership? It's a larger change, but ultimately what's needed to do this without repeating this pattern across services that need this kind of enrichment.

yahyafakhroji:
@scotwells agreed, resolving references in the activity service is the right end state. The annotation only exists to get the email into the audit payload. Opened milo-os/activity#249.

Can we land this as a stop-gap? https://github.com/datum-cloud/cloud-portal/issues/1507 is live and enrichment has no timeline. Once it ships, the webhook, annotation, controller and infra patch all come out together.

@JoseSzycho your design stands as-is. The annotation goes away when enrichment lands.

JoseSzycho:
@yahyafakhroji did you consider what @scotwells mentioned of using a control plane events  ?

This would allow you to get rid of the controller and this extra complexity

yahyafakhroji:
@JoseSzycho I did look at it. Two things stopped me.

There's no GroupMembership controller on main, so events need a new controller too. It emits instead of patching. The webhook, annotation and infra route do go away, so it's still less machinery overall.

The blocker is the actor. Event rules take it from `reportingController`, so the line reads as the controller instead of the person who added the member. Milo doesn't record the human creator anywhere a controller could read it. Deletion also needs a finalizer, since the delete audit entry covers it today.

@scotwells moved to enrichment after I raised these (milo-os/activity#249), which keeps the actor from the audit log.

So the question is whether the line has to name the person. If yes, the options are landing this as a stop-gap or waiting for milo-os/activity#249. Either works for me, I just need a call.

PR #792, opened 2026-09-16
fix(quota): don't claim project note quota for platform-scoped subjects

kevwilliams:
## Summary

Creating a `Note` against a `Contact` — or any note whose subject isn't project-scoped — failed with a 403 and an unhelpful message:

```
notes.notes.miloapis.com "note-xxxxx" is forbidden: Something went wrong while
checking your quota for this request. Please try again — if this keeps
happening, contact support.
```

This blocks the staff-portal Notes tab on the Contact detail page, and would hit any client creating a note for a platform-scoped subject.

## Root cause

The quota admission plugin infers a claim's `ConsumerRef` from which control-plane path prefix the request arrived on (`pkg/quota/admission/plugin.go:852-872`), via the filters in `pkg/server/filters/{projects,organizations}.go`. A note about a `Contact` in `milo-system` reaches neither `/projects/{id}/control-plane/` nor `/organizations/{org}/control-plane/`, so **both** `ProjectID(ctx)` and `OrganizationID(ctx)` are empty and the claim ends up with no consumer at all.

`notes-per-project` requires a `Project` consumer, so the empty consumer mismatches. That mismatch should be a clean denial, but `IsClaimingResourceAllowed` (`pkg/quota/validation/resourcetype.go:241-249`) returns a raw Go error instead of `(false, nil, nil)`. It becomes a `field.InternalError`, can't be classified as a known claim failure, and falls through to the generic "internal error" bucket and its fallback message.

Worth noting the API already disagreed with itself here: `Note` and `ClusterNote` are annotated `discovery.miloapis.com/parent-contexts=Platform`, while the only registration for them enforces a `Project` consumer, and the claim policies fired on *every* note regardless of subject.

## What this changes

Notes about platform-scoped subjects have no owning project, so they shouldn't have been claiming project note quota to begin with. Both claim policies now carry a `spec.trigger.constraints` expression excluding platform-scoped subject API groups. When a constraint is false the plugin returns `nil` (`plugin.go:528-536`) — no claim is created and the note is admitted.

Also re-enables `test/crm/note-contact-lifecycle`, which covers `Note`→`Contact` and `ClusterNote`→`User` and was marked `skip: true`. That skip is why this shipped broken.

## Deliberately not fixed here

- **The error classification.** `resourcetype.go:241-249` should return a denial, not an error. It's a one-liner and independent of this unblock, but it's what turns the next occurrence of this class from "please try again" into an accurate message. Left for a separate PR.
- **Metering.** Notes about platform-scoped subjects are now unmetered. These paths are staff-only and RBAC-gated, but whether they should be metered — and against which consumer, given a `Contact` has no owning org — is an open product question. A fuller design doc covering subject-derived consumer resolution is drafted separately.
- **The denylist is a maintenance burden.** A new platform-scoped subject in some other API group will hit the original 403. The error-classification fix above is what makes that case self-diagnosing.

## Test plan

- [x] CEL expression validated and evaluated against the real engine (`NewCELEngine` / `ValidateConstraints` / `EvaluateConditions`): returns `false` for `notification.miloapis.com`, `iam.miloapis.com`, `resourcemanager.miloapis.com`; `true` for `networking.datumapis.com`, the Domain case that must keep being metered.
- [x] All three YAML files parse; `constraints` lands under `spec.trigger` as a sibling of `resource`.
- [x] `task test:unit` passes, including all six quota packages.
- [ ] **`task test:end-to-end` — not run locally, no cluster available. This is the main thing CI needs to confirm**, since the unskipped test has never run against a live API server.


JoseSzycho:
@gaghan430 FYI. You have been waiting weeks for this fix

JoseSzycho (commented):
@kevwilliams what prevents a user to create an unlimited number of Notes for their projects or user?

kevwilliams:
@JoseSzycho Short answer is nothing other than the notes-creator/notes-editor/notes-admin being a manually given role as far as I can tell. Im open to suggestions on how we can do it better.

kevwilliams:
Settled on User for the consumer. `spec.creatorRef` is already on every Note, and the mutating webhook sets it from the authenticated user rather than trusting the client, so it's a server-controlled key — can't be spoofed into someone else's bucket. Ordering works out too: the quota plugin is a validating plugin, so it runs after the mutating webhooks and `creatorRef` is always populated by the time the claim template renders.

So the shape is: a second registration for `notes.miloapis.com/notes` with `consumerType: iam.miloapis.com/User`, and a claim policy for platform-scoped subjects whose template sets `consumerRef.name` from `{{trigger.spec.creatorRef.name}}`. The plugin already honors a template-set ConsumerRef and only falls back to path inference when it's blank, so this is almost entirely config.

Two things whoever picks this up needs to know:

- **A grant has to land first.** Bucket limits come only from active ResourceGrants and start at 0, so shipping the registration without a GrantCreationPolicy on User creation would deny every note instead of capping them. Registration and grant policy go in together or not at all.
- **Per-creator means the busiest staff member hits the wall first.** That's the honest tradeoff of keying on the user rather than a global bucket. Probably fine with a generous limit, but it's a real property, not an accident.

Proposal: leave the constraint in this PR as the unblock so staff-portal isn't waiting on the above, and do the User-consumer work as a follow-up that removes the constraint again. Shout if you'd rather hold this one and do it in a single pass.

kevwilliams:
Couple of corrections to the above after digging further — the direction holds, two of the specifics were wrong.

**It can't be a second registration on `notes.miloapis.com/notes`.** The registration cache is keyed by resource type alone (`v.cache[resourceType]`, `resourcetype.go:354`) with no duplicate guard, so two active registrations on the same type would silently clobber each other — last reconcile wins — and deactivating either would evict the other's entry. The User-consumer one needs its own type string, e.g. `notes.miloapis.com/user-notes`.

**The constraint in this PR doesn't get removed by the follow-up.** It stays; the new policy just takes the complementary set. The two policies partition the space between them rather than one superseding the other.

**Also: grants aren't missing, I was looking in the wrong repo.** They live in `datum`, not here — `config/services/notes.miloapis.com/quota/grant-policies/default-project-grant-policy.yaml` grants 100 notes per project on Project creation, with a sibling core policy for secrets/configmaps. So project notes are capped at 100 and working; nothing is broken. Milo ships registrations + claim policies, datum ships the grants.

One nice thing that falls out of that: the datum policies set `target.parentContext` because a project note's claim lands in the project's own control plane, so the grant has to be written there too. A User-consumer grant doesn't need that — notes on contacts are created in the core control plane, and Users live there as well, so the policy omits `parentContext` entirely and writes locally. That's also the branch that sets an owner reference on the trigger (`grant_creation_executor.go:380`), so a user's grant gets garbage-collected with the user for free — something the project-scoped policies don't get.

So the follow-up spans two repos: milo gets the new registration + claim policies, datum gets a User-triggered GrantCreationPolicy with no parentContext.

scotwells (approved):
I think this is fine for now. We should look at updating the quota system to not claim from a consumer when there's no consumer in the request.

mattdjenkinson (approved):

kevwilliams:
@scotwells I was thinking similar but @JoseSzycho's question about how would we count them etc came to mind as an issue if we werent claiming at all, and it would fail open and allow things so I may be mistaken but think this is "better" because it 403s. I'm definitely open to revisiting it I just might be missing something.

scotwells:
@kevwilliams I'm not too concerned with tracking platform scoped resources in the quota system. Consumers of the platform will never create platform level resources.

kevwilliams:
ok so maybe we dont need #793 at all then because that really only exists to meter the contact notes to the author.

scotwells:
Given that only platform users would be creating notes on contacts, I don't think it's necessary.

PR #793, opened 2026-09-16
feat(quota): meter notes on platform-scoped subjects against their author

kevwilliams:
Follow-up to #792 (stacked on it — review that one first).

## Summary

#792 stopped notes on contacts, users and organizations from claiming project quota, which unblocked them but left them unmetered. This meters them against the note's author instead, so staff notes are counted and limits become possible later.

- New `notes-per-user` registration, consumer `iam.miloapis.com/User`
- Claim policies for platform-scoped subjects, resolving the consumer from `spec.creatorRef`
- The project and user policies carry complementary constraints and partition note subjects between them

`creatorRef` is set by the mutating webhook from the authenticated user, so it can't be spoofed into someone else's quota, and the quota plugin validates after mutating webhooks run, so it's always populated by the time the claim template renders.

## Two things worth knowing

**It needs a new resource type, not a second registration on `notes.miloapis.com/notes`.** The registration cache is keyed by resource type alone with no duplicate guard, so two registrations on one type would silently clobber each other.

**⚠️ Blocked on datum.** Bucket limits come only from active grants and start at 0, so merging this before a `GrantCreationPolicy` for `notes.miloapis.com/user-notes` exists in datum would deny contact notes again — re-breaking exactly what #792 fixed. The chainsaw tests supply the grant directly, so CI here passes either way; production would not. **Please don't merge until the datum policy is in place.**

The datum policy should omit `parentContext` — notes on contacts are created in the core control plane where Users also live, so the grant is written locally, and that's the branch that sets an owner reference, so a user's grant is garbage-collected with the user.

## Test plan

- [x] Unit test asserts the two policies partition subject groups — exactly one fires for any group, checked against the shipped YAML so a future subject kind can't fall through both or match both
- [x] Unit test renders the user policy's claim template and asserts the consumer resolves to the note's author and the request targets `user-notes`
- [x] `kustomize build config/services` renders all three new objects
- [x] `task test:unit` passes
- [ ] New `test/quota/user-notes-quota` chainsaw test — not run locally, no cluster. Asserts a contact note creates a claim consuming the author's quota, that it's granted, and that no project-scoped claim is created.
- [ ] `test/crm/note-contact-lifecycle` gains a grant fixture, since its notes now consume author quota. Also unverified locally.


kevwilliams:
We decided this is not needed right now in #792

